PR #3, opened 2018-04-13
Allow the Y axes to be removed

evancharlton:
By using the following configuration options on the <DataProvider>, the
y axes will not be rendered (the series still will be).

config: {
  yAxis {
    display: 'NONE',
  }
}

Supported values are 'NONE' and 'ALL'. Any other values will behave as
if 'ALL' was specified.

### DIFF
--- a/src/components/ChartContainer/index.js
+++ b/src/components/ChartContainer/index.js
@@ -33,7 +33,7 @@ class ChartContainer extends Component {
   getChartWidth = () => {
     const {
       width,
-      yAxis: { width: yAxisWidth },
+      yAxis: { width: yAxisWidth, display: yAxisDisplayMode },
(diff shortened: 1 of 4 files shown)


src/components/ChartContainer/index.js
```diff
@@ -95,6 +92,9 @@ class ChartContainer extends Component {
       .domain(subDomain)
       .range([0, chartWidth]);
     const children = React.Children.map(this.props.children, child => {
+      if (!child) {
```
martincognite: Can this ever be the case?

evancharlton: Yes. When I remove the child dynamically, this.props.children[1] === false.

I have no idea why.

PR #11, opened 2018-04-19
Allow the lines to have adjustable thickness

evancharlton:
Give <DataProvider> a new prop, `strokeWidths`, which will control the
width of the rendered lines (in pixels).

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "demo:build": "nwb build-react-component src/index.js"
   },
   "dependencies": {
+    "antd": "^3.4.1",
     "bluebird": "^3.5.1",
     "d3": "^4.12.0",
     "immutable": "^3.8.2",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -23,6 +23,7 @@
     "demo:build": "nwb build-react-component src/index.js"
   },
   "dependencies": {
+    "antd": "^3.4.1",
```
martincognite: Could this potentially clash with the dependencies of the components that's using this? This is a general question, not only about antd, I just know that specific one is in fact in operational-intelligence as well.

evancharlton: I don't know anything about declaring dependencies in NodeJS. Should I remove it and use a different component instead?

martincognite: Just leave it, I'll look into it.

PR #13, opened 2018-04-20
Fix the warnings on the storybook

evancharlton:
LineChart / DataProvider are expecting an object, not an array.

### DIFF
--- a/stories/index.js
+++ b/stories/index.js
@@ -605,7 +605,7 @@ storiesOf('DataProvider', module)
           ...baseConfig,
         };
 
-        state = { showTimeline: true, strokeWidths: [] };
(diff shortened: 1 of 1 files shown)


stories/index.js
```diff
@@ -635,9 +635,9 @@ storiesOf('DataProvider', module)
                   value={1}
                   min={1}
                   max={10}
-                  step={1}
+                  step={0.25}
                   onChange={value => {
-                    const copy = [...strokeWidths];
+                    const copy = { ...strokeWidths };
```
martincognite: Sidenote; you can do
`{ ...strokeWidths, [key]: value }` as well.

PR #14, opened 2018-04-23
Fixing a bug with multiple onClick handlers.

martincognite:


martincognite:
Temporary solution. We need to rethink how the onClick handler should be interacted with.

evancharlton:
Filed https://github.com/cognitedata/griff-react/issues/15 to track this

PR #16, opened 2018-04-25
Refactoring, moving should component update into the render components.

martincognite:


martincognite:
I'll hold off merging this a bit, want to do some manual testing into opint to make sure we're not breaking any core functionality. I left it quite open - not stopping to many changes. Should be improved over time as we hit performance errors.

PR #20, opened 2018-05-01
Create an encapsulated Series object

evancharlton:
As data enters the DataProvider, convert it into an instance of a Series
object. This will contain all of the metadata as well as the functions
that callers will use down the line, like scale calculations.

This also allows the callers to have the metadata live directly with the
series object passed in from the loader, instead of as a series of
disconnected metadata objects (eg: hiddenSeries, colors, strokeWidths).
Backwards compatibility is maintained by mixing the metadata arrays in
the DataProvider, with the series object's metadata taking precedence.

Finally, create an AxisCollection object and extract a dedicaetd YAxis
object. These maintain single-responsibility behavior, where they are
only responsible for rendering their level of data. This gives us much
more manageable code in LineChart and Axis. It also allows for an
<AxisCollection> to be added as a sibling of <LineChart>, but this has
not yet been implemented because it causes issues with calculating the
chart width once the axes become dynamically shown and hidden.

### DIFF
--- a/src/components/DataProvider/index.js
+++ b/src/components/DataProvider/index.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { uniq, isEqual } from 'lodash';
+import Series from '../../data';
 
 export default class DataProvider extends Component {
   state = {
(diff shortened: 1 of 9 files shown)


src/components/DataProvider/index.js
```diff
@@ -144,18 +145,57 @@ export default class DataProvider extends Component {
...
+          series.staticDomain = yAxis.staticDomain[key];
+        }
+        if (!series.calculateDomain && yAxis.calculateDomain) {
+          series.calculateDomain = yAxis.calculateDomain;
+        }
+        if (!series.xAccessor && xAxis.accessor) {
+          series.xAccessor = xAxis.accessor;
+        }
+        if (!series.yAccessor && yAxis.accessor) {
+          series.yAccessor = yAxis.accessor;
```
martincognite: Should we support a default accessor?

evancharlton: In the storybook, we have `baseConfig.yAxis` and `baseConfig.xAxis` specify their default accessors. If we decide to support default ones, that should be a separate PR, IMO

PR #31, opened 2018-05-04
Added SURGE_TOKEN as kube secret, Jenkins needs it to deploy the stor…

martincognite:
…ybook.


evancharlton:
Do we have any restrictions on only allowing this to happen if the PR comes from someone in the cognitedata organization? We don't want to allow random people to run code on our PR server.

PR #36, opened 2018-05-11
Add eslint and prettier support

valerii-cognite:
Added cognite-react-scripts in order to have eslint and prettier capabilities in the project. Also added `yarn build:watch` script which basically listens for the source code changes and runs `yarn build`, which is pretty handy when repository is used as a link.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,3 @@
+/demo/dist
+node_modules/
+coverage/
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,29 @@
+{
+  "extends": ["airbnb", "prettier", "prettier/react"],
+  "plugins": ["react", "prettier", "jest"],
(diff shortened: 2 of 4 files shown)


.eslintignore
```diff
@@ -0,0 +1,3 @@
+node_modules/
+/demo/dist
+coverage/
```
evancharlton: Please add a trailing newline

.eslintrc.json
```diff
@@ -0,0 +1,29 @@
...
+    "import/no-extraneous-dependencies": 0,
+    "import/no-unresolved": 0,
+    "import/extensions": 0,
+    "import/prefer-default-export": 0,
+    "prettier/prettier": [
+      "error",
+      { "singleQuote": true, "trailingComma": "es5" }
+    ]
+  }
+}
```
evancharlton: Same thing here -- add a trailing newline

PR #39, opened 2018-05-12
Removed all eslint warnings

evancharlton:
This change is a massive cleanup -- it adds eslint to the dependencies
and goes through and fixes all detected warnings. This mainly includes
defining propTypes and defaultProps, and writing custom prop validation.

Code which is not ours is excluded via the .eslintignore file.


evancharlton:
Note: I'm fine not merging this -- the most valuable code here is probably `Validators.js`, followed by the cleanup to `stories`. Other than that, it was just a fun exercise. If it's going to help with the 0.2.0 refactor, then we can merge it. If not, we can skip it.

evancharlton:
I'll redo this on 0.2.0 some day.

PR #40, opened 2018-05-14
Live loading

valerii-cognite:
The chart live loading is now available. It utilizes updates of xScale through the ChartContainer series prop. I've also added a DataProvider `shift` property, if it's true, the data will be shifted from the left on a new point adding.

### DIFF
--- a/src/components/ChartContainer/index.js
+++ b/src/components/ChartContainer/index.js
@@ -47,12 +47,46 @@ class ChartContainer extends Component {
     return width - yAxisWidth * visibleAxesCount - margin.left - margin.right;
   };
 
+  getCalcSubDomain(series, hiddenSeries, xAxis, subDomain) {
+    let min = Number.MIN_SAFE_INTEGER;
+    let max = Number.MAX_SAFE_INTEGER;
+    Object.keys(series)
+      .filter(key => !hiddenSeries[key])
+      .forEach(key => {
+        const serie = series[key];
+        const xAccessor = serie.xAccessor || xAxis.accessor;
+        const data = serie.data.filter(d => {
+          const value = xAccessor(d);
+          return subDomain[0] < value && value < subDomain[1];
+        });
+        const xDomainSerie = this.calculateDomainFromData(data, xAccessor);
+        min = Math.max(min, xDomainSerie[0]);
+        max = Math.min(max, xDomainSerie[1]);
+      });
+    return min === Number.MIN_SAFE_INTEGER || max === Number.MAX_SAFE_INTEGER
+      ? subDomain
+      : [min, max];
+  }
+
+  calculateDomainFromData = (data, yAccessor) => {
+    const extent = d3.extent(data, yAccessor);
+    const diff = extent[1] - extent[0];
+    if (Math.abs(diff) < 1e-3) {
+      return [1 / 2 * extent[0], 3 / 2 * extent[0]];
+    }
+    return [extent[0] - diff * 0.025, extent[1] + diff * 0.025];
+  };
+
   subDomainChanged = domain => {
     this.props.subDomainChanged(domain);
   };
 
   updateTransformation = s => {
-    const { margin, series, yAxis: { width: yAxisWidth } } = this.props;
+    const {
+      margin,
+      series,
+      yAxis: { width: yAxisWidth },
+    } = this.props;
     const nSeries = Object.keys(series).length;
     const width = this.getChartWidth();
     this.setState({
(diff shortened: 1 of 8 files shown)


src/components/ChartContainer/index.js
```diff
@@ -47,12 +47,45 @@ class ChartContainer extends Component {
...
 
+  getCalcSubDomain(series, hiddenSeries, xAxis, subDomain) {
+    let min = Number.MAX_SAFE_INTEGER;
+    let max = 0;
+    Object.keys(series)
+      .filter(key => !hiddenSeries[key])
+      .forEach(key => {
+        const serie = series[key];
+        const xAccessor = serie.xAccessor || xAxis.accessor;
+        const data = serie.data.filter(
```
evancharlton: If we know that the data is ordered, we could be clever and stop looking once `xAccessor(d) > subDomain[1]`

Similarly, we should probably only do `xAccessor(d)` once

src/components/ChartContainer/index.js
```diff
@@ -47,12 +47,45 @@ class ChartContainer extends Component {
...
+          d => xAccessor(d) > subDomain[0] && xAccessor(d) < subDomain[1]
+        );
+        const xDomainSerie = this.calculateDomainFromData(data, xAccessor);
+        if (xDomainSerie[0] < min) min = xDomainSerie[0];
```
evancharlton: `Math.max(min, xDomainSerie[0])`

src/components/ChartContainer/index.js
```diff
@@ -47,12 +47,45 @@ class ChartContainer extends Component {
...
+        if (xDomainSerie[1] > max) max = xDomainSerie[1];
```
evancharlton: `Math.min(max, xDomainSerie[1])`

src/components/ChartContainer/index.js
```diff
@@ -47,12 +47,45 @@ class ChartContainer extends Component {
     return width - yAxisWidth * visibleAxesCount - margin.left - margin.right;
   };
 
+  getCalcSubDomain(series, hiddenSeries, xAxis, subDomain) {
+    let min = Number.MAX_SAFE_INTEGER;
+    let max = 0;
```
evancharlton: This should be `Number.MIN_SAFE_INTEGER` to handle negative x values

src/components/DataProvider/index.js
```diff
@@ -157,6 +154,32 @@ export default class DataProvider extends Component {
...
+      // follow most right subdomain
+      if (this.state.subDomain[1] === this.state.domain[1]) {
+        update.subDomain = [this.state.subDomain[0], domainMax];
+      }
+
+      // shift serie data array if a new item has been added to the end
+      if (this.props.shift) {
+        Object.keys(series).forEach(key => {
+          if (series[key].data.length > this.state.series[key].data.length) {
+            series[key].data.shift();
```
evancharlton: Don't we need to to call `shift()` the number of times as however many elements were added?

src/components/ChartContainer/index.js
valerii-cognite: Isn't is that code readability over small performance optimization? Consider the following code
```
const data = [];
for (let i = 0; i < serie.data.length; i += 1) {
  const d = serie.data[i];
  const value = xAccessor(d);

  if (subDomain[0] < value && value < subDomain[1]) {
    data.push(d);
  }

  if (value > subDomain[1]) {
    break;
  }
}
```
vs
```
const data = serie.data.filter(d => {
  const value = xAccessor(d);
  return subDomain[0] < value && value < subDomain[1];
});
```

evancharlton: \*shrug\*

Considering we bill this as a high-performance charting library, we should consider performance optimizations, especially if they're happening on `render()`.

With that being said: I'm okay leaving this one and revisiting it once we're in the optimization phase.

src/components/DataProvider/index.js
valerii-cognite: Good catch

valerii-cognite:
It seems it was a better idea just to add another commit instead of force push in order to keep history, isn't it? In anyway the only question which is still open is in the first review comment.

src/components/ChartContainer/index.js
valerii-cognite: That's a good idea and thank you for pointing out.

src/components/DataProvider/index.js
```diff
@@ -157,6 +154,37 @@ export default class DataProvider extends Component {
...
+      }
+
+      // shift serie data array if a new item has been added to the end
+      if (this.props.shift) {
+        Object.keys(series).forEach(key => {
+          const newSerieLength = series[key].data.length;
+          const oldSerieLength = this.state.series[key].data.length;
+          if (newSerieLength > oldSerieLength) {
+            const lengthDiff = newSerieLength - oldSerieLength;
+            for (let i = 0; i < lengthDiff; i += 1) {
```
evancharlton: suggestion: You could do `for (let i = oldSerieLength; i < newSerieLength; i += 1)` instead of calculating the diff.

If you think that's worse readability, then I'm indifferent.

valerii-cognite: agree

valerii-cognite:
Close in favor of https://github.com/cognitedata/griff-react/pull/51

PR #42, opened 2018-05-16
Add support for min/max area

evancharlton:
If y0Accessor and y1Accessor are specified with a series, draw an area
surrounding the line. This allows for uncertainty to be drawn if
desired.


evancharlton:
Small change to calculate the domain based on the min/max instead of the data. Please take another look!

PR #45, opened 2018-05-16
Add ruler support

valerii-cognite:


valerii-cognite:
Done, I've also added touch support

stories/index.js
```diff
@@ -714,4 +707,43 @@ storiesOf('DataProvider', module)
...
+            1: 'red',
+            2: 'green',
+            3: 'blue',
+          }}
+        >
+          <ChartContainer>
+            <LineChart
+              heightPct={0.8}
+              ruler={{
+                xLabel: point =>
```
evancharlton: Aren't these flipped?

valerii-cognite: You are right

valerii-cognite:
> It would be nice to have some logic in there about not overlapping the labels, but that can come as a separate PR.
Agree, as well as yAxis overlapping, I will make another PR for that

> It's also a little weird to snap the ruler to the datapoints -- it's kind of jarring to move around.
Yes, but in between the points we do not have any values, so there is not obvious what to show. BTW the Highcharts ruler works the same way.

martincognite:
Agree with the overlapping, this is a sought after feature in opint as well.

However, I was thinking that you might want to implement this client-side instead of in this library. That way we can get away with a lot of the styling decisions and give a lot more control. I think that would at least be the way for 0.2.0 which is giving away more of the styling control to the user.

We should extend `onMouseMove` to also register touch events (or register new handlers) and just like we do in the current `onMouseMove` expose the x and y coordinates`. This way people can either use the return values as legends (as we currently do in opint), tooltips -- or even both if that makes sense.

I would merge this if you need it in opsup (and potentially rewrite it in a 0.2.x patch) anyway.

valerii-cognite:
> However, I was thinking that you might want to implement this client-side instead of in this library. That way we can get away with a lot of the styling decisions and give a lot more control. I think that would at least be the way for 0.2.0 which is giving away more of the styling control to the user.

I'm absolutely fine to move such a functionality into a separate layer (i.e. opsup, opint, etc) to get rid of any styling decisions on the library level. But I have some concerns about the ruler, isn't it suppose to be a build-in must have feature since almost all the libraries which are already out there does have it as a feature? If that is not the case and we are not trying to compete with an "everything-in" libraries we are fine.

> I would merge this if you need it in opsup (and potentially rewrite it in a 0.2.x patch) anyway.

We already have it as a prerelease in a separate brunch, so we are fine. If you take that decision of moving the functionality, I will rewrite ruler to meet the 0.2.x requirements.

valerii-cognite:
Added overlapping fixes

valerii-cognite:
Close in favor of https://github.com/cognitedata/griff-react/pull/52

PR #47, opened 2018-05-25
0.2.0 ruler loading

valerii-cognite:
The PR handles the following issues:
1. Handling new series
2. Handling new domain
3. Adds the ruler functionality inside library

### DIFF
--- a/src/components/DataProvider/index.js
+++ b/src/components/DataProvider/index.js
@@ -44,22 +44,29 @@ export default class DataProvider extends Component {
   }
 
   componentDidUpdate(prevProps) {
-    const { series: newSeries } = this.props;
-    const { series: oldSeries } = prevProps;
+    const { series: newSeries, baseDomain: newDomain } = this.props;
+    const { series: oldSeries, baseDomain: oldDomain } = prevProps;
 
     for (let i = 0; i < newSeries.length; i += 1) {
       if (!oldSeries.find(s => s.id === newSeries[i].id)) {
-        this.fetchData('SERIES_UPDATED');
-        break;
+        return this.fetchData('SERIES_UPDATED');
       }
     }
 
     for (let j = 0; j < oldSeries.length; j += 1) {
       if (!newSeries.find(s => s.id === oldSeries[j].id)) {
-        this.fetchData('SERIES_UPDATED');
-        break;
+        return this.fetchData('SERIES_UPDATED');
       }
     }
+
+    if (
+      newDomain &&
+      oldDomain &&
+      (newDomain[0] !== oldDomain[0] || newDomain[1] !== oldDomain[1])
+    ) {
+      return this.fetchData('NEW_DOMAIN');
+    }
+    return 1;
   }
 
   componentWillUnmount() {
@@ -84,7 +91,8 @@ export default class DataProvider extends Component {
 
   fetchData = async reason => {
     const { baseDomain, pointsPerSeries, defaultLoader } = this.props;
-    const { subDomain } = this.state;
+    const subDomain =
+      reason === 'NEW_DOMAIN' ? baseDomain : this.state.subDomain;
     const newLoaderConfig = {};
     const seriesObjects = this.getSeriesObjects();
     const enhancedSeries = await Promise.map(seriesObjects, async s => {
--- a/src/components/DataProvider/index.js
+++ b/src/components/DataProvider/index.js
@@ -30,9 +30,8 @@ const deleteUndefinedFromObject = obj => {
 export default class DataProvider extends Component {
   state = {
     subDomain: this.props.baseDomain,
-    loaderConfig: {},
     contextSeries: {},
-    yDomains: {},
+    enhancedSeries: [],
   };
 
   async componentDidMount() {
@@ -44,6 +43,25 @@ export default class DataProvider extends Component {
     }
   }
 
+  componentDidUpdate(prevProps) {
+    const { series: newSeries } = this.props;
+    const { series: oldSeries } = prevProps;
+
+    for (let i = 0; i < newSeries.length; i += 1) {
+      if (!oldSeries.find(s => s.id === newSeries[i].id)) {
+        this.fetchData('SERIES_UPDATED');
+        break;
+      }
+    }
+
+    for (let j = 0; j < oldSeries.length; j += 1) {
+      if (!newSeries.find(s => s.id === oldSeries[j].id)) {
+        this.fetchData('SERIES_UPDATED');
+        break;
+      }
+    }
+  }
+
   componentWillUnmount() {
     clearInterval(this.fetchInterval);
     this.unmounted = true;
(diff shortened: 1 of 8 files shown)


src/components/Ruler/index.js
```diff
@@ -0,0 +1,96 @@
...
+  points: PropTypes.arrayOf(pointPropType).isRequired,
+  ruler: rulerPropType.isRequired,
+};
+
+const labelHeight = 24;
+
+const calculateY = (points, y) => {
+  const pointsObject = {};
+
+  const pointsSorted = [...points].sort((a, b) => {
```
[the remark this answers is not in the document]
valerii-cognite: This will not handle strings, please see [this](https://jsfiddle.net/ue91go1d/). 
But I can go with localeCompare if you prefer.

src/components/Ruler/index.js
```diff
@@ -0,0 +1,96 @@
...
+      prevPoint = pointsSorted[i].y;
+    }
+    pointsObject[pointsSorted[i].name] = prevPoint;
+  }
+
+  return pointsObject;
+};
+
+const Ruler = ({ ruler, points, height, width }) => {
+  const yLabelHeight = height - labelHeight - 5;
```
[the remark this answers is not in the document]
valerii-cognite: This is just a tooltip margin

src/components/Ruler/index.js
```diff
@@ -0,0 +1,96 @@
...
+  const pointsObject = calculateY(points, yLabelHeight);
```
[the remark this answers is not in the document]
valerii-cognite: please see a new commit

src/components/Ruler/index.js
```diff
@@ -0,0 +1,96 @@
...
+const propTypes = {
+  width: PropTypes.number.isRequired,
+  height: PropTypes.number.isRequired,
+  points: PropTypes.arrayOf(pointPropType).isRequired,
+  ruler: rulerPropType.isRequired,
+};
+
+const labelHeight = 24;
+
+const calculateY = (points, y) => {
```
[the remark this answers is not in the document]
valerii-cognite: This is a position of yTooltipPosition, please see a new commit

src/components/Ruler/index.js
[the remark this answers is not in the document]
evancharlton: Bizarre. How would `localeCompare` work with non-string y values, though? (for that matter, I don't understand why we would have string y values?)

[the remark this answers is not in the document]
valerii-cognite: Right, it should be only a number. I'll fix that

src/components/Ruler/index.js
```diff
@@ -44,8 +40,9 @@ const calculateY = (points, y) => {
 };
 
 const Ruler = ({ ruler, points, height, width }) => {
-  const yLabelHeight = height - labelHeight - 5;
-  const pointsObject = calculateY(points, yLabelHeight);
+  const yTooltipMargin = 5;
+  const yTooltipPosition = height - labelHeight - yTooltipMargin;
```
evancharlton: I'm still not understanding why this is called `yTooltipPosition`? It seems like `pointsObject` is actually the tooltip positions, right? Is this just the available region that is able to be used when positioning tooltips?

valerii-cognite: It is just a position of timestamp tooltip which suppose to stick to the most bottom position. Please check the Ruler story.

src/components/DataProvider/index.js
```diff
@@ -44,31 +43,59 @@ export default class DataProvider extends Component {
...
+    const { series: oldSeries, baseDomain: oldDomain } = prevProps;
+
+    for (let i = 0; i < newSeries.length; i += 1) {
+      if (!oldSeries.find(s => s.id === newSeries[i].id)) {
+        return this.fetchData('SERIES_UPDATED');
+      }
+    }
+
+    for (let j = 0; j < oldSeries.length; j += 1) {
+      if (!newSeries.find(s => s.id === oldSeries[j].id)) {
```
martincognite: Existing series is deleted by default, the only thing we need to do is ask for data if there's a new serie coming in (even if it's toggled).

src/components/InteractionLayer/index.js
```diff
@@ -199,15 +167,70 @@ class InteractionLayer extends React.Component {
     }
   };
 
+  processMouseMove = (
```
martincognite: Why does this need to have `props` as input? It has access to `this.props` itself.

src/components/InteractionLayer/index.js
```diff
@@ -199,15 +167,70 @@ class InteractionLayer extends React.Component {
...
+    }
+
+    if (onMouseMove) {
+      onMouseMove({ points: newPoints, xpos, ypos });
+    }
+  };
+
   zoomed = () => {
+    if (this.props.ruler) {
+      this.processMouseMove(this.state.touchX, this.state.touchY, this.props);
```
martincognite: Very nice. Didn't think of this.

src/components/DataProvider/index.js
```diff
@@ -78,33 +105,32 @@ export default class DataProvider extends Component {
...
+        oldSerie: s,
         reason,
       });
+      const data =
+        Array.isArray(loaderReturn.data) && loaderReturn.data.length
+          ? loaderReturn.data
+          : [];
       return {
-        data: [],
+        ...s,
```
martincognite: This is breaking a lot of the other stories. The reason why `this.getSeriesObjects` are used in `render` as opposed to in `fetchData` is to allow updates to propagate when the `series` props change (`zoomable`, `hidden` etc). Now -- nothing will happen until `fetchData` is called again.

src/components/DataProvider/index.js
```diff
@@ -78,33 +105,32 @@ export default class DataProvider extends Component {
         baseDomain,
         subDomain,
         pointsPerSeries,
-        oldSeries: s,
+        oldSerie: s,
```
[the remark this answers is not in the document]
evancharlton: "series" :-)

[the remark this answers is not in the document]
valerii-cognite: we need somehow to specify our expectations, do we need to use seriesArray instead? :)

src/components/DataProvider/index.js
valerii-cognite: If we had 2 series and now we have only 1 we need to update the chart somehow. Do we need just to filter seriesArray in that case?

src/components/InteractionLayer/index.js
valerii-cognite: Good point, will fix that

src/components/DataProvider/index.js
martincognite: Longer explanation;

The thought behind `getSeriesObjects` is to merge the two different series configs. One set of properties comes from the `series` prop of `DataProvider` (`id`, `loaders`, `zoomable` etc) and the other set of properties comes from the loader itself (`data`, `step`, `drawPoints`). These two configs can be updated independently, so merging them together in `render` ensures that whichever one of those changed, they get merged and shipped through the `context`. If we update the state only in `fetchData` then the `series` prop sent through `DataProvider` will not reflect any changes until `fetchData` is called and `enhancedSeries` is re-built.

src/components/InteractionLayer/index.js
```diff
@@ -6,10 +6,12 @@ import ScalerContext from '../../context/Scaler';
 import { createXScale, createYScale } from '../../utils/scale-helpers';
 import { seriesPropType, annotationPropType } from '../../utils/proptypes';
 import Annotation from '../Annotation';
+import Ruler from '../Ruler';
 
 class InteractionLayer extends React.Component {
   static propTypes = {
     crosshair: PropTypes.bool,
+    ruler: PropTypes.objectOf(PropTypes.func).isRequired,
```
evancharlton: Could you use `shape` for this?
```
ruler: PropTypes.shape({xLabel: PropTypes.func.isRequired, yLabel: PropTypes.func.isRequired}),
```

valerii-cognite: Already used that [here](https://github.com/cognitedata/griff-react/pull/52). I close this one in in favor of that PR.

PR #51, opened 2018-05-25
[WIP] Live loading in 0.2.0 first try

valerii-cognite:


martincognite:
Should we have the loader handling the slicing of the data on `INTERVAL` updates? In that case, you can fetch the latest (one or more) and add it to the end of the existing data array. Then it's also up to the loader/you to remove points from the front of you want a sliding window -- or you can leave them be to expand the `baseDomain`.

Another approach would be to simply shift the `baseDomain` from `[t0, t1]` to `[t0 + update_inverval, t1 + update_interval]` in `fetchData` as we know how often we're asking for updates. This would create a  sliding window. In which case we could make an exception for the `INTERVAL` reason and append of to the existing data in `fetchData` while shifting the subdomain. If no points were added -- the window will still update.

valerii-cognite:
It seems like the second approach is easier to implement involving more transparent logic to the code. That's also meant we are not adhering series data anymore, but should we really care about it? It is also more clear to a user that live loading is happening, but should he lose the point to the left even if no new point came to the right? We can discuss more on Monday and I think this approach can also take place.

Issue #24, opened 2018-05-03
Check if onMouseMove before doing all the work


martincognite:
Title pretty much explains it.

evancharlton:
This was fixed with #52

PR #52, opened 2018-05-25
0.2.0 Ruler functionality

valerii-cognite:
This is a child PR of https://github.com/cognitedata/griff-react/pull/47

### DIFF
--- a/src/components/InteractionLayer/index.js
+++ b/src/components/InteractionLayer/index.js
@@ -4,12 +4,18 @@ import * as d3 from 'd3';
 import isEqual from 'lodash.isequal';
 import ScalerContext from '../../context/Scaler';
 import { createXScale, createYScale } from '../../utils/scale-helpers';
-import { seriesPropType, annotationPropType } from '../../utils/proptypes';
+import {
+  seriesPropType,
+  annotationPropType,
+  rulerPropType,
+} from '../../utils/proptypes';
 import Annotation from '../Annotation';
+import Ruler from '../Ruler';
 
 class InteractionLayer extends React.Component {
   static propTypes = {
     crosshair: PropTypes.bool,
+    ruler: rulerPropType,
     height: PropTypes.number.isRequired,
     onClick: PropTypes.func,
     onClickAnnotation: PropTypes.func,
@@ -34,6 +40,11 @@ class InteractionLayer extends React.Component {
     updateXTransformation: () => {},
     series: [],
     zoomable: true,
+    ruler: {
+      visible: false,
+      xLabel: () => {},
+      yLabel: () => {},
+    },
   };
 
   state = {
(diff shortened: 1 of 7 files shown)


src/components/InteractionLayer/index.js
```diff
@@ -106,50 +104,17 @@ class InteractionLayer extends React.Component {
...
-            id: s.id,
-            timestamp: ts,
-            value,
-            x: xScale(ts),
-            y: yScale(value),
-          });
-        } else {
-          output.points.push({ id: s.id });
-        }
+    if (onMouseMove || ruler) {
```
evancharlton: This fixes #24 -- don't forget to put that in your commit!

src/components/Ruler/index.js
```diff
@@ -0,0 +1,93 @@
...
+    }
+    pointsObject[pointsSorted[i].name] = prevPoint;
+  }
+
+  return pointsObject;
+};
+
+const Ruler = ({ ruler, points, height, width }) => {
+  const xTooltipMargin = 5;
+  const xTooltipPosition = height - labelHeight - xTooltipMargin;
```
evancharlton: Now it's called `xTooltipPosition`? I'm even more confused.

What does this value actually represent? As far as I can tell, it's just a point 29px (`labelHeight + xTooltipMargin`) off the bottom (`height - 29`) of the chart, but I don't know why, or what it's used for?

evancharlton:
I get this in the storybook: 
```
warning.js:33 Warning: Failed prop type: The prop `ruler` is marked as required in `InteractionLayer`, but its value is `undefined`.
```

src/components/InteractionLayer/index.js
valerii-cognite: Good catch

src/components/Ruler/index.js
valerii-cognite: ok, just added a comment

valerii-cognite:
Missed proper propTypes, fixed

src/components/RulerTooltip/index.js
```diff
@@ -0,0 +1,78 @@
...
+};
+
+class RulerTooltip extends Component {
+  state = {
+    textWidth: 0,
+  };
+
+  componentDidMount() {
+    const textBBox = this.textRef.getBBox();
+    // eslint-disable-next-line react/no-did-mount-set-state
```
evancharlton: Why are we able to ignore this?

martincognite: Is this a `setState` in did/will mount discussion?

valerii-cognite: It seems the second render is not an issue [anymore](https://github.com/airbnb/javascript/issues/684#issuecomment-381674576), so airbnb want to turn this error off. The other problem we have I didn't think about before is that we can't really know in advance the text width has been already set in componentDidMount hook, but as far as it is synchronous operation we can run setState right after all the sync operations finish by wrapping it in an immediate setTimeout, i.e.
```
setTimeout(() =>
  this.setState({
    textWidth: textBBox.width,
  })
);
```

PR #53, opened 2018-05-25
Added live loading story

valerii-cognite:


stories/index.js
```diff
@@ -383,4 +391,18 @@ storiesOf('LineChart', module)
       }
       return <ZoomToggle />;
     })
+  )
+  .add(
+    'Live loading',
+    withInfo()(() => (
+      <DataProvider
+        defaultLoader={staticLoader}
+        baseDomain={staticBaseDomain}
```
martincognite: How do we make sure `baseDomain` reflects the outer range of the zoom? Technically only the datapoints in the range of `baseDomain` will be shown.

We could use `baseDomain` as a baseline and keep that in state -- and when `updateInterval` is used, and when `fetchData(id, 'INTERVAL')` is being ran we can modify the `baseDomain` in the state of `DataProvider`.

This implies that when `baseDomain` from props is changed -- it will reset.

martincognite: Saw this one before your first actual crack at an implementation -- nevermind.

PR #55, opened 2018-05-25
Add/remove series

martincognite:
When adding or removing series, the lifecycle currently broke. By changing the pattern in `fetchData` that does its operation on every single series (assuming they're all bundles) to calling `fetchData` for every series we have a lot more flexibility for tuning each individual series.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,8 +29,9 @@
     "bluebird": "^3.5.1",
     "d3": "^4.12.0",
     "immutable": "^3.8.2",
-    "np": "^2.20.1",
     "lodash.isequal": "^4.5.0",
+    "np": "^2.20.1",
+    "react-select": "^1.2.1",
     "react-sizeme": "^2.4.3",
     "styled-components": "^3.2.6"
   },
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "lodash.isequal": "^4.5.0",
     "np": "^2.20.1",
     "react-select": "^1.2.1",
-    "react-sizeme": "^2.4.3",
+    "react-sizeme": "2.4.3",
     "styled-components": "^3.2.6"
   },
   "peerDependencies": {
--- a/src/components/DataProvider/index.js
+++ b/src/components/DataProvider/index.js
@@ -100,6 +100,8 @@ export default class DataProvider extends Component {
     await Promise.map(newSeries, async ({ id }) => {
       await this.fetchData(id, 'MOUNTED');
       if (!isEqual(subDomain, baseDomain)) {
+        // The series got added when zoomed in,
+        // Need to also fetch a higher-granularity version on mount
         await this.fetchData(id, 'UPDATE_SUBDOMAIN');
       }
     });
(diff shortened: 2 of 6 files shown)


stories/index.js
```diff
@@ -430,4 +432,44 @@ storiesOf('LineChart', module)
...
+    withInfo()(() => {
+      const options = [
+        { value: '1', label: 'steelblue' },
+        { value: '2', label: 'maroon' },
+        { value: '3', label: 'red' },
+      ];
+      // eslint-disable-next-line
+      class EnableDisableSeries extends React.Component {
+        state = {
+          series: options,
```
valerii-cognite: if instead of options specify array of a single series and try to add data through combobox, an error occurs:
`Uncaught TypeError: Array.prototype.map called on null or undefined`

src/components/DataProvider/index.js
```diff
@@ -36,20 +37,40 @@ export default class DataProvider extends Component {
...
       this.fetchInterval = setInterval(() => {
-        this.fetchData('INTERVAL');
+        Promise.map(this.props.series, s => this.fetchData(s.id, 'INTERVAL'));
       }, this.props.updateInterval);
     }
   }
 
+  async componentDidUpdate(prevProps, prevState) {
+    const { series: prevSeries } = prevState;
+    if (!prevSeries) {
```
valerii-cognite: It seems like an issue here, since series is not defined in state, the if statement will be everytime `true` and the logic below will never be addressed.

src/components/DataProvider/index.js
```diff
@@ -36,20 +37,40 @@ export default class DataProvider extends Component {
...
+      return;
+    }
+    const { series, baseDomain } = this.props;
+    const { subDomain } = this.state;
+    const prevSeriesKeys = {};
+    prevSeries.forEach(p => {
+      prevSeriesKeys[p.id] = true;
+    });
```
valerii-cognite: `prevSeriesKeys` never used further

src/components/DataProvider/index.js
```diff
@@ -61,52 +82,69 @@ export default class DataProvider extends Component {
...
+      );
+    }
+    return {
+      data: [],
+      ...deleteUndefinedFromObject(series),
+      ...deleteUndefinedFromObject(loaderConfig[id]),
+      xAccessor: series.xAccessor || xAccessor,
+      yAccessor: series.yAccessor || yAccessor,
+      yDomain: series.yDomain || yDomains[id],
+    };
```
valerii-cognite: yeah, we can just unify this object return logic with `getSeriesObjects`

package.json
```diff
@@ -29,9 +29,10 @@
     "bluebird": "^3.5.1",
     "d3": "^4.12.0",
     "immutable": "^3.8.2",
-    "np": "^2.20.1",
     "lodash.isequal": "^4.5.0",
-    "react-sizeme": "^2.4.3",
+    "np": "^2.20.1",
+    "react-select": "^1.2.1",
```
valerii-cognite: it seems like an overhead to keep react-select inside the library as long as we use it only in stories, or webpack will handle that somehow?

martincognite: Webpack will not bundle it. It will start at src/index.js and not include what's exclusively used in stories.

src/components/DataProvider/index.js
```diff
@@ -35,21 +36,81 @@ export default class DataProvider extends Component {
     yDomains: {},
   };
 
+  static getDerivedStateFromProps(nextProps, prevState) {
+    // Check if one of the series got removed from props
+    // If so, delete the respective keys in contextSeries and loaderconfig
+    // This is important so we don't cache the vales if it gets readded later
```
evancharlton: Good catch

src/components/DataProvider/index.js
```diff
@@ -35,21 +36,81 @@ export default class DataProvider extends Component {
...
+    });
+  }
+
   componentWillUnmount() {
     clearInterval(this.fetchInterval);
     this.unmounted = true;
   }
 
-  getSeriesObjects() {
+  getSeriesObjects = () => {
```
evancharlton: This is sort of a meta question, but why is `foo = () => {` preferred over `foo() {` ?

valerii-cognite: Using arrow function notation is just the way of saving the lexical scope in `this` in order to call it from another class which became available as a part of [this proposal](https://github.com/tc39/proposal-class-fields). Consider you pass function as a prop to the child class inside of which you want to access parent's scope (i.e. to use props, setState, etc). If just to pass it as is, you will lose the scope, since `this` will be related to a child class. Consider the following:

```
const Child = ({handleClick}) => 
  <button onClick={handleClick}>Button</button>

class Parent extends Component {
  handleClick() {
    console.log(this.props);
  }

  render() {
    return 
  }
}
```


There are lot's of ways to workaround this behaviour, one of which is method arrow notation, another instead of passing method, wrap it in an arrow function and call (`() => this.method()`), another one is use `.bind()`, etc.

BTW there is actually no need to wrap `getSeriesObjects` here since it is called only inside the same class.

martincognite: I agree -- however `() => method()` will create a new function every time. There is no need here -- but I like the consistency.

martincognite:
This is ready to merge on my side. @valerii-cognite ?

PR #57, opened 2018-05-30
[WIP] Added support for live loading.

martincognite:
This enables live loading from the data loader.

The `loader` should handle the `INTERVAL` reason and return the existing series + the new datapoint(s). If an `updateInterval` prop is sent to the `DataProvider`, it will change it's internal `baseDomain` to `[t0 + updateInterval, t1 + updateInterval]` and propagate this down. The `Scaler` will look for changes in the `baseDomain`, and if the trailing edge of the current `subDomain` is equal to the trailing edge of the `baseDomain`, it will change it's internal subDomain. Each component will make sure that zooming/brushing will be seamless in this transition.

## Short comings
- Expanding the initial baseDomain will not trigger a lifecycle method. Not really sure how this should be handled, for now I would recommend having a key on the `DataProvider` that includes `${baseDomain[0]}_${baseDomain[1]}` to force a full refresh of the chart when the prop changes. This will trigger the necessary lifecycle methods (`MOUNTED` for the loaders, updating y domains etc). This should be inspected more.

- When your current subDomain goes out of scope (`subDomain[0] < baseDomain[0]`) is doesn't to anything clever other than keep the current view. If you start to zoom you will be snapped back to `subDomain[0] === baseDomain[0]`.

## Discussions

- The loader can chose to omit the points from the old series where `point.timestamp < baseDomain[0]` before returning from the `INTERVAL` clause -- or better yet leave one point to be able to draw a line from the first point and out to the left. These points could be washed away automatically by the `DataProvider`, but I prefer not having `DataProvider` mutate the data and leave that responsibility to the application.

- There could be a separate prop indicating whether it should be a sliding or expanding window, that is if `newBaseDomain = [t0 + dt, t0 + dt]` or `newBaseDomain = [t0, t0 + dt]`. This is currently not implemented.

### DIFF
--- a/src/components/DataProvider/index.js
+++ b/src/components/DataProvider/index.js
@@ -31,6 +31,7 @@ const deleteUndefinedFromObject = obj => {
 export default class DataProvider extends Component {
   state = {
     subDomain: this.props.baseDomain,
+    baseDomain: this.props.baseDomain,
     loaderConfig: {},
     contextSeries: {},
     yDomains: {},
@@ -71,10 +72,19 @@ export default class DataProvider extends Component {
 
   async componentDidMount() {
     await Promise.map(this.props.series, s => this.fetchData(s.id, 'MOUNTED'));
-    if (this.props.updateInterval) {
+    const { updateInterval } = this.props;
+    if (updateInterval) {
       this.fetchInterval = setInterval(() => {
-        Promise.map(this.props.series, s => this.fetchData(s.id, 'INTERVAL'));
-      }, this.props.updateInterval);
+        const { baseDomain } = this.state;
+        this.setState(
+          { baseDomain: baseDomain.map(d => d + updateInterval) },
+          () => {
+            Promise.map(this.props.series, s =>
+              this.fetchData(s.id, 'INTERVAL')
+            );
+          }
+        );
+      }, updateInterval);
     }
   }
 
@@ -85,8 +95,8 @@ export default class DataProvider extends Component {
     if (!prevSeries) {
       return;
     }
-    const { series, baseDomain } = this.props;
-    const { subDomain } = this.state;
+    const { series } = this.props;
+    const { subDomain, baseDomain } = this.state;
 
     const currentSeriesKeys = {};
     series.forEach(s => {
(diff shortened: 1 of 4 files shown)


src/components/DataProvider/index.js
```diff
@@ -105,6 +115,14 @@ export default class DataProvider extends Component {
         await this.fetchData(id, 'UPDATE_SUBDOMAIN');
       }
     });
+
+    // Check if basedomain changed in props -- if so reset state.
+    if (!isEqual(this.props.baseDomain, prevProps.baseDomain)) {
```
martincognite: This is touched on in the discussion -- this is a fallback for the props (will update based on feedback/discussion).

src/components/DataProvider/index.js
```diff
@@ -105,6 +115,14 @@ export default class DataProvider extends Component {
         await this.fetchData(id, 'UPDATE_SUBDOMAIN');
       }
     });
+
+    // Check if basedomain changed in props -- if so reset state.
```
evancharlton: Should this cancel the live loading? And if `updateInterval` changed, then `fetchInterval` needs to be canceled and recreated

martincognite: Good catch -- fixed.

valerii-cognite:
> The loader can chose to omit the points from the old series where point.timestamp < baseDomain[0] before returning from the INTERVAL clause

It's a bit unclear to me, why the LineChart keeps the left points as a default, but ContextChart doesn't, so I can see it smoothly moving inside the bottom window. Don't we need to implement the same behaviour for the LineChart and if shift prop is passed as false to DataProvider just not update the left baseDomain inside startUpdateInterval method?

martincognite:
Yes we can -- hence the discussion part. My reasoning here is that if you're zoomed in somewhere you don't want that lose that.

martincognite:
I actually think it would, it's getting messy to call `fetchData` in the same way multiple places (`await Promise.map(this.props.series, ({ id }) => this.fetchData(id, reason))`).

Let's move it to another PR.

PR #58, opened 2018-05-30
Implementation of custom brush

martincognite:
D3’s brush actually updates the DOM - which means that draws to the DOM happens without react knowing about it, hence not being able to be clever about the virtual dom diffing (I think). D3s zoom however only attaches an event listener to a given object and leaves the rest up to the user (great). Also, the zoom events are a lot more complicated to replicate as it handles all kinds of touch events. D3s brush is also a lot more complicated than necessary as it’s built to handle xy zoom which we’re not using.

## TODO:
Debug this with touch events, if that will ever become a use case. Currently — Operation Support is not using the context chart afaik, so there’s no rush there.

### DIFF
--- a/src/components/Brush/index.js
+++ b/src/components/Brush/index.js
@@ -3,12 +3,12 @@ import PropTypes from 'prop-types';
 
 class Brush extends React.Component {
   state = {
-    isDraggingOverlay: false,
-    isDraggingHandleWest: false,
-    isDraggingHandleEast: false,
-    isDraggingSelection: false,
     dragStartOverlay: null,
     dragStartSelection: null,
+    isDraggingHandleEast: false,
+    isDraggingHandleWest: false,
+    isDraggingOverlay: false,
+    isDraggingSelection: false,
   };
 
   componentDidMount() {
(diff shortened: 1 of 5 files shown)


src/components/Brush/index.js
```diff
@@ -1,81 +1,207 @@
...
+
   componentDidMount() {
-    const { width, height } = this.props;
-    this.brush = d3
-      .brushX()
-      .extent([[0, 0], [width, height]])
-      .on('brush end', this.didBrush);
-    this.selection = d3.select(this.brushNode);
-    this.selection.call(this.brush);
+    window.addEventListener('mouseup', this.onMouseUp);
```
evancharlton: Do you want to check `zoomable` before we register these?

src/components/Brush/index.js
```diff
@@ -1,81 +1,207 @@
...
+          onMouseDown={this.onMouseDownSelection}
         />
         <rect
-          className="handle handle--e"
-          cursor="ew-resize"
-          x={selection[0] - 5}
-          y="-3"
-          width={10}
-          height={86}
+          className="handle handle--west"
```
evancharlton: `handle--west` here vs `handle-east` below

src/components/Brush/index.js
martincognite: I don't think the cost of registering them (considering we're throwing early if it's not zoomable) is any high. No strong opinion.

evancharlton: Yeah me neither. Leave it in.

martincognite:
It's mimicking the existing behaviour, there should be no difference except for performance.

PR #73, opened 2018-06-05
Export AxisDisplayMode

evancharlton:
This needs to be exported so that clients can use this.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-73.surge.sh

PR #74, opened 2018-06-06
Ticks amount depends on the width

valerii-cognite:
In opsup we have a problem of overlapping xAxis labels for smaller devices. This is basically because griff relies on d3 [`axis.ticks()`](https://github.com/d3/d3-3.x-api-reference/blob/master/SVG-Axes.md#ticks) which default number is 10 no mater width the chart has. It seems there is no any proper solution over there, so I came up with idea of just passing a number of ticks depending on the width. Better ideas of course are welcome.

### DIFF
--- a/src/components/XAxis.js
+++ b/src/components/XAxis.js
@@ -56,7 +56,12 @@ class Axis extends Component {
     const tickSizeOuter = axis.tickSizeOuter();
     const tickSizeInner = axis.tickSizeInner();
     const tickPadding = axis.tickPadding();
-    const values = scale.ticks();
+    // In order to reduce label overlapping for smaller devices
+    // we want to adjust amount of ticks depending on width.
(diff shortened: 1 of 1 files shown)


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-74.surge.sh

martincognite:
Great idea. How about minimum 2 ticks?

valerii-cognite:
It seems d3 [`axis.ticks()`](https://github.com/d3/d3-3.x-api-reference/blob/master/SVG-Axes.md#ticks) is not guarantees it will show the exact amount of ticks passed as argument, so I will stick with 1 as a default so to make sure at least one label is show on the screen.

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-74.surge.sh

PR #76, opened 2018-06-11
Move the collapsed axis to the outside

evancharlton:
Render the placeholder CollapsedAxis to the outside of the
AxisCollection instead of the inside. It shouldn't be between the real
axes and the chart contents.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-76.surge.sh

PR #78, opened 2018-06-13
Calculate ticks of yAxis based on height

valerii-cognite:
Same fix as for xAxis but for yAxis. Personally I don't like those almost `random` numbers (100 and 50) , better ideas of course are welcome.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-78.surge.sh

PR #82, opened 2018-06-18
Don't check for ruler; check for ruler.visible

evancharlton:
`this.props.ruler` will always be truthy; we need to check whether it is
visible or not.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-82.surge.sh

PR #87, opened 2018-06-20
Change the axis events to be per-axis

evancharlton:
Instead of triggering axis mouse events on the `<AxisCollection>`, set
them on each axis, and add the series ID to the callback.

Mouse events on the collapsed axis are passed with a series ID of
'collapsed'.

### DIFF
--- a/src/components/AxisCollection/index.js
+++ b/src/components/AxisCollection/index.js
@@ -36,10 +36,14 @@ const defaultProps = {
 };
 
 class AxisCollection extends React.Component {
+  onAxisMouseEnter = seriesId => e => this.props.onMouseEnter(e, seriesId);
+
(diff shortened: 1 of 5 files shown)


src/components/AxisCollection/index.js
```diff
@@ -83,27 +89,22 @@ class AxisCollection extends React.Component {
...
-    const axes = this.renderAllAxes();
+    const axes = this.renderAllVisibleAxes();
     return (
-      <svg
-        width={width}
-        height={height}
-        onMouseEnter={onMouseEnter}
-        onMouseLeave={onMouseLeave}
-        onFocus={onMouseEnter}
-        onBlur={onMouseLeave}
```
valerii-cognite: Just curious, why do you think we don't need those `onFocus` and `onBlur` events to trigger?

PR #97, opened 2018-06-21
Fix "checker is not a function" warning

evancharlton:
`PropTypes.shape` cannot take in a class, so define the shape we
actually need instead.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-97.surge.sh

PR #98, opened 2018-06-21
Don't allow `<rect>` to be negative

evancharlton:
Due to floating-point math, we can wind up with a negative width,
which results in a warning in the console.

The rect should be zero-width anyway, so bound them to 0.t


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-98.surge.sh

PR #103, opened 2018-06-21
Make externalSubDomain not required

valerii-cognite:


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-103.surge.sh

valerii-cognite:
The value is actually `null` since we have a default value for `subDomain` on `DataProvider` level and I don't think it can hurt us.

PR #104, opened 2018-06-24
Check for `area` existence before using it

evancharlton:
If for some reason we get a mouseup event before we get a mousedown
event, the `area` variable can be null. In this case, we should simply
silently fail instead of crashing.

### DIFF
--- a/src/components/InteractionLayer/index.js
+++ b/src/components/InteractionLayer/index.js
@@ -198,7 +198,7 @@ class InteractionLayer extends React.Component {
     }, 50);
     if (onAreaDefined) {
       const { area } = this.state;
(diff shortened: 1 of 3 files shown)


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-104.surge.sh

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-104.surge.sh

martincognite:
Quite the edge case. I assume this is related to https://github.com/cognitedata/operational-intelligence/issues/567 ?

evancharlton:
Yep.

PR #106, opened 2018-06-25
Extract an idPropType

evancharlton:
Consolidate all of the `id` props into a single proptype definition --
which can be a number or a string.

Additionally, don't require `Area` to have a start and end property
since this won't happen when the area first pops into existence.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-106.surge.sh

PR #107, opened 2018-06-27
Fixing domain calculation when 0 is the only value present.

martincognite:


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-107.surge.sh

PR #121, opened 2018-07-30
Added opening animation to lines in a chart

danlevings:
Added a prop to the LineChart called 'openWithAnimation'.

This should animate both the line and the area smoothly

### DIFF
--- a/src/components/Line/index.js
+++ b/src/components/Line/index.js
@@ -2,103 +2,162 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import * as d3 from 'd3';
 
-const Line = ({
-  data,
-  xAccessor,
-  yAccessor,
-  y0Accessor,
-  y1Accessor,
-  xScale,
-  yScale,
-  color,
-  step,
-  hidden,
-  drawPoints,
-  strokeWidth,
-  clipPath,
-}) => {
-  let line;
-  let area;
-  // HTML has an issue with drawing points somewhere in the 30-35M range.
-  // There's no point in drawing pixels more than 30k pixels outside of the range
-  // so this hack will work for a while.
-  // Without this, when zoomed far enough in the line will disappear.
-  const boundedSeries = value => Math.min(Math.max(value, -30000), 30000);
-  if (step) {
-    line = d3
-      .line()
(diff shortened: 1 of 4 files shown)


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-121.surge.sh

src/components/Line/index.js
```diff
@@ -2,103 +2,162 @@ import React from 'react';
...
+    return (
+      <g clipPath={`url(#${clipPath})`}>
+        {area && (
+          <path
+            className="line-area"
+            d={area(data)}
+            style={{
+              stroke: 'none',
+              strokeWidth: `${strokeWidth}px`,
+              transition: 'clip-path 2s ease',
```
evancharlton: Don't we only want to do this if animations are set?

src/components/LineChart/index.js
```diff
@@ -86,6 +89,7 @@ const defaultProps = {
   areas: [],
   onAreaDefined: null,
   onAreaClicked: null,
+  openWithAnimation: false,
```
evancharlton: What about `animationDuration` being a number? 0/null/undefined/false = nothing

stories/index.js
```diff
@@ -119,6 +119,18 @@ storiesOf('LineChart', module)
       </DataProvider>
     ))
   )
+  .add(
+    'Basic with animation',
```
evancharlton: Could you add an animated min/max example, too?

danlevings:
Changed prop to 'openingAnimationDuration', improved stories to customise speed, fixed weird lines to the right - enjoy!

danlevings:
Technically, thats right - you could probably set this up to be outside of griff entirely.
Although imo it feels like it would be part of a new griff add-on for animations and other flashy effects

danlevings:
So is it decided that we should scrap this PR and move it inside opint?

evancharlton:
I would very much prefer this live in the client app for the time being -- but if there's anything you need (css classes, eg) to make it happen, please send a PR or file an issue.

danlevings:
This will be implemented externally from griff (https://github.com/cognitedata/operational-intelligence/pull/1078) - closing

PR #127, opened 2018-08-02
Show storybook info inline

evancharlton:
Move the withInfo call to the global storybook config so that stories
are rendered with their props and source code inline.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-127.surge.sh

PR #130, opened 2018-08-12
Support configurable axis placements

### LINKED ISSUE #89: Support left-side axes
Y axes should be able to be rendered on the left side as well as the right. In fact, it should support showing both at once.

evancharlton:
Add a new `yAxisPlacement` property to both LineChart and series. This
value is one of the defined `AxisPlacement` constants, and it controls
where the respective axes/axis live/s.

It has the following behaviors:
  - UNSPECIFIED
     - Same as not using the property, and the default value (RIGHT) is used.
-  RIGHT
    - Default behavior -- the y-axes are rendered to the right of the line chart.
-  LEFT
    - Axes are rendered to the left of the line chart; series are mirrored so that the proximity to the chart is preserved.
-  BOTH
    - Axes are rendered on both sides of the chart. These are kept in sync with each other.

If the series objects specify this option, then they will be placed as
specified, overriding the chart's value if necessary.

Fixes #89

### DIFF
--- a/src/components/AxisCollection/CollapsedAxis.js
+++ b/src/components/AxisCollection/CollapsedAxis.js
@@ -3,7 +3,8 @@ import * as d3 from 'd3';
 import PropTypes from 'prop-types';
 import isEqual from 'lodash.isequal';
 import { createYScale } from '../../utils/scale-helpers';
-import { singleSeriePropType } from '../../utils/proptypes';
+import { singleSeriePropType, axisPlacementType } from '../../utils/proptypes';
+import AxisPlacement from '../LineChart/AxisPlacement';
 
 export default class CollapsedAxis extends Component {
   static propTypes = {
@@ -25,6 +26,7 @@ export default class CollapsedAxis extends Component {
     color: PropTypes.string,
     onMouseEnter: PropTypes.func,
     onMouseLeave: PropTypes.func,
+    yAxisPlacement: axisPlacementType,
   };
 
   static defaultProps = {
@@ -36,6 +38,7 @@ export default class CollapsedAxis extends Component {
     offsetx: 0,
     onMouseEnter: null,
     onMouseLeave: null,
+    yAxisPlacement: AxisPlacement.RIGHT,
   };
 
   componentWillMount() {
@@ -57,6 +60,40 @@ export default class CollapsedAxis extends Component {
       }
     }
   }
+
+  getPath = ({ offsetx, strokeWidth, tickSizeOuter, range }) => {
+    const { yAxisPlacement, width } = this.props;
+    switch (yAxisPlacement) {
+      case AxisPlacement.LEFT:
+        return [
+          // Move to this (x,y); start drawing
+          `M ${width - offsetx} ${strokeWidth}`,
+          // Draw a horizontal line to the left
+          `h ${tickSizeOuter - strokeWidth}`,
+          // Draw a vertical line from top to bottom
+          `v ${range[0] - strokeWidth * 2}`,
+          // Finish with another horizontal line
+          `h -${tickSizeOuter - strokeWidth / 2}`,
+        ].join(' ');
+      case AxisPlacement.BOTH:
+        throw new Error(
+          'BOTH is not a valid option for CollapsedAxis -- please specify RIGHT or LEFT'
+        );
+      case AxisPlacement.RIGHT:
+      case AxisPlacement.UNSPECIFIED:
+      default:
+        return [
+          // Move to this (x,y); start drawing
+          `M ${offsetx} ${strokeWidth}`,
+          // Draw a horizontal line to the left
+          `h -${tickSizeOuter - strokeWidth}`,
+          // Draw a vertical line from top to bottom
+          `v ${range[0] - strokeWidth * 2}`,
+          // Finish with another horizontal line
+          `h ${tickSizeOuter - strokeWidth / 2}`,
+        ].join(' ');
+    }
+  };
 
   syncZoomingState = () => {
     if (this.props.zoomable) {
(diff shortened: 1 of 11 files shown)


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-130.surge.sh

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-130.surge.sh

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-130.surge.sh

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-130.surge.sh

src/components/AxisCollection/CollapsedAxis.js
```diff
@@ -36,6 +38,7 @@ export default class CollapsedAxis extends Component {
     offsetx: 0,
     onMouseEnter: null,
     onMouseLeave: null,
+    yAxisPlacement: AxisPlacement.RIGHT,
```
martincognite: ANY REASON WE HAVE THIS TO THE RIGHT AS DEFAULT????

src/components/AxisCollection/YAxis.js
```diff
@@ -59,6 +65,100 @@ export default class YAxis extends Component {
...
+        return [
+          `M${width - tickSizeOuter},${range[0] - strokeWidth}`,
+          `H${width - strokeWidth}`,
+          `V${range[1]}`,
+          `H${width - tickSizeOuter}`,
+        ].join('');
+      case AxisPlacement.BOTH:
+        throw new Error(
+          'BOTH is not a valid option for YAxis -- please specify RIGHT or LEFT'
+        );
```
valerii-cognite: What is the scenario when this will throw an error and why?

src/components/AxisCollection/CollapsedAxis.js
```diff
@@ -58,6 +61,40 @@ export default class CollapsedAxis extends Component {
...
+          `h ${tickSizeOuter - strokeWidth}`,
+          // Draw a vertical line from top to bottom
+          `v ${range[0] - strokeWidth * 2}`,
+          // Finish with another horizontal line
+          `h -${tickSizeOuter - strokeWidth / 2}`,
+        ].join(' ');
+      case AxisPlacement.BOTH:
+        throw new Error(
+          'BOTH is not a valid option for CollapsedAxis -- please specify RIGHT or LEFT'
+        );
```
valerii-cognite: What is the scenario when this will throw an error and why?

evancharlton: In case we (griff developers) just blindly pass in `yAxisPlacement` without saying *which* side it's on.

For example, if the user sets the default to be `BOTH`, and then the developer just passes that into `CollapsedAxis`, rather than making a `LEFT` version and a `RIGHT` version. This will help us catch it in the storybook.

src/components/AxisCollection/CollapsedAxis.js
evancharlton: JUST KEEPING IT CONSISTENT!!!11!

src/components/AxisCollection/index.js
```diff
@@ -47,45 +50,93 @@ class AxisCollection extends React.Component {
...
-                width={yAxisWidth}
-                updateYTransformation={updateYTransformation}
-                yTransformation={yTransformations[s.id]}
-                onMouseEnter={this.onAxisMouseEnter(s.id)}
-                onMouseLeave={this.onAxisMouseLeave(s.id)}
-              />
-            );
-          })
+        filteredSeries.filter(s => s.collectionId === undefined).map(s => {
+          axisOffsetX += yAxisWidth;
```
martincognite: Will this push the first axis `yAxisWidth` pixels in as well?

src/components/LineChart/index.js
```diff
@@ -111,84 +116,116 @@ class LineChartComponent extends Component {
...
+      .concat(series.filter(s => s.collectionId === undefined))
+      .concat(collections)
+      .reduce(
+        (acc, item) => {
+          if (item.yAxisPlacement) {
+            acc[item.yAxisPlacement] = (acc[item.yAxisPlacement] || 0) + 1;
+          }
+          return acc;
+        },
+        { [yAxisPlacement]: 1 }
```
martincognite: I don't understand the initial value here, I'm sure it's on my end though.

src/components/AxisCollection/index.js
evancharlton: Yeah, but note that on L101 I change it to subtract the yAxisWidth to start.

src/components/LineChart/index.js
evancharlton: It's to account for the default. Example:
```
<LineChart
  yAxisPlacement={AxisPlacement.BOTH}
  series={[
    {id: 1, yAxisPlacement: AxisPlacement.LEFT},
    {id: 2},
  ]}
/>
```

In this case, you'll have `2` on both axes, and `1` only on the left. So you want the output of this reducer to be:
```
{
  BOTH: 1,
  LEFT: 1,
}
```

Does that clarify things? I'll put in a comment to explain the gist of this in the code.

evancharlton: Actually, scratch that. In the process of adding a comment, I realized that it's not actually totally correct -- it would double-count the default.

I reworked it and added a story to verify the behavior -- please take a look!

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-130.surge.sh

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-130.surge.sh

PR #131, opened 2018-08-12
Add support for grid lines

### LINKED ISSUE #113: Background grid lines
It would be great if Griff supported background grid lines. I don't really have much of a specification for this other than maybe the following options:

 - Horizontal-only
 - Vertical-only
 - Both horizontal and vertical
 - Statically set (eg, every 50 px)
 - Locked to the values (eg, every 1.0 units)

evancharlton:
Expose a `grid` property on the LineChart which controls the grid line
specification. This object is documented in `proptypes.js`, but here's a
summary:

The `y` property of the grid specification controls how horizontal
lines behave (they're distributed along the Y values). These can either
be rendered statically or linked to one or more series. If there are
links, then the grid lines will be transformed just like the y-axis tick
marks as the data is scaled.

The `x` property of the grid specification controls how vertical lines
behave (they're distributed along the X values). Similarly, these can
either be rendered statically or linked to the time value. There is no
series linkage necessary for this, because the x-axis is independent of
the series being plotted.

Fixes #113

### DIFF
--- a/src/components/GridLines/index.js
+++ b/src/components/GridLines/index.js
@@ -8,33 +8,22 @@ const propTypes = {
   grid: gridPropType,
   height: PropTypes.number.isRequired,
   width: PropTypes.number.isRequired,
-  opacity: PropTypes.number,
-  color: PropTypes.string,
-  strokeWidth: PropTypes.number,
   series: seriesPropType.isRequired,
   subDomain: PropTypes.arrayOf(PropTypes.number).isRequired,
 };
 
 const defaultProps = {
   grid: null,
-  opacity: 0.4,
-  color: '#000',
-  strokeWidth: 1,
 };
+
+const DEFAULT_COLOR = '#666';
+const DEFAULT_OPACITY = 0.6;
+const DEFAULT_STROKE_WIDTH = 1;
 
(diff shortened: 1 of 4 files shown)


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-131.surge.sh

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-131.surge.sh

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-131.surge.sh

valerii-cognite:
That's a cool feature to have!

It seems that 
- top-level props for grid are not taken into account
- on a GridLines component level there are some other top-level props, i.e. opacity, color, which operated as defaults, so I would better move them outside of the class to constants, since them are a bit misleading to me.

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-131.surge.sh

evancharlton:
* Top-level grid props are now respected for all configurations (with stories to prove it)
 * Default prop values extracted to named constants instead of hiding them in the `props` object.

PR #136, opened 2018-08-19
Don't use a default end for Area

evancharlton:
If there is a default provided, then an initial area (from [0,0]) will
be drawn when when mousedown happens, until the mousemove is registered.

Default it to null so that we can detect this state and abort the
rendering.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-136.surge.sh

PR #140, opened 2018-08-29
Allow min/max area to be styled with CSS

evancharlton:
Expose all of the properties necessary so that client applications can
choose their own styles for the min/max area shade, if desired.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-140.surge.sh

PR #141, opened 2018-08-29
Fix publishing of tags

evancharlton:
Jenkins wasn't logging in with its GitHub credentials, so the tags could
not be pushed properly. Fix this.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-141.surge.sh

PR #142, opened 2018-08-29
Another try at pushing GitHub tags

evancharlton:
I think `yarn publish` won't even do the commit if it doesn't know who it is (duh). Gotta init git first -- and in the same subshell as the commit. Hence, all one script.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-142.surge.sh

PR #143, opened 2018-08-29
Don't pass in default series & collections

evancharlton:
YAxis should not be given default series or collections, as that can
only be misleading later on. Instead, they should be null, and the
correct behavior will be explicit during the execution.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-143.surge.sh

PR #146, opened 2018-08-30
Allow the point width to be controlled

evancharlton:
Much like the calling application can control the stroke width, they can
also control the width of the points being rendered. This is done with a
new pointWidth property, which can be set on either the LineChart or a
series object.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-146.surge.sh

PR #148, opened 2018-09-01
Add point-hovering capabilities

evancharlton:
Add a callback function to the series object: onPointHover. This is
called whenever the point is hovered (or un-hovered) to allow the
application to respond accordingly.

This change also puts the points on top of the InteractionLayer for
scatterplots. Without this change, the hovering functionality is blocked
by the InteractionLayer. However, zooming is blocked when on top of a
point.

### DIFF
--- a/src/components/Points/index.js
+++ b/src/components/Points/index.js
@@ -14,6 +14,8 @@ const propTypes = {
   xScale: scaleFuncPropType.isRequired,
   yScale: scaleFuncPropType.isRequired,
   color: PropTypes.string.isRequired,
+  // (nullableData, e) => void
+  onPointHover: PropTypes.func,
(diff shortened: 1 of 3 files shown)


evancharlton:
@martincognite - if you know of a way to allow the points to sit on top of the interaction layer without blocking the zoom functionality, I'm **all ears**.

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-148.surge.sh

martincognite:
I don't know too much about this, but it should be possible to propagate the events both up and down the stack properly.

evancharlton:
I'm trying to figure that one out before I merge an obvious regression.

PR #154, opened 2018-09-07
Added onZoomXAxis to LineChart API,

danlevings:
and added className to annotations for easy front end searching

### DIFF
--- a/src/components/Annotation/index.js
+++ b/src/components/Annotation/index.js
@@ -4,6 +4,7 @@ import { annotationShape } from '../../utils/proptypes';
 const Annotation = ({ data, xScale, height, color, fillOpacity, id }) => (
   <rect
     key={id}
+    className={`id-${id}`}
     x={xScale(data[0])}
     y={0}
     height={height}
--- a/src/components/LineChart/index.js
+++ b/src/components/LineChart/index.js
@@ -39,6 +39,7 @@ const propTypes = {
   onClickAnnotation: PropTypes.func,
   // event => void
   onDoubleClick: PropTypes.func,
+  onZoomXAxis: PropTypes.func,
   subDomain: PropTypes.arrayOf(PropTypes.number),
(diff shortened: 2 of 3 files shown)


src/components/Annotation/index.js
```diff
@@ -4,6 +4,7 @@ import { annotationShape } from '../../utils/proptypes';
 const Annotation = ({ data, xScale, height, color, fillOpacity, id }) => (
   <rect
     key={id}
+    className={`id-${id}`}
```
evancharlton: `id-foo` seems like not a very descriptive (or unique) class name. What do you think about something like:
```
className={`griff-annotation griff-annotation-${id}`}
```

src/components/LineChart/index.js
```diff
@@ -39,6 +39,7 @@ const propTypes = {
   onClickAnnotation: PropTypes.func,
   // event => void
   onDoubleClick: PropTypes.func,
+  onZoomXAxis: PropTypes.func,
```
evancharlton: Please document what arguments this can expect to have. It looks like it doesn't get any, but that's pretty surprising to me. Could we pass back the new subDomain (x-axis) or the transformation object (or both?) or something like that?

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-154.surge.sh

src/components/LineChart/index.js
```diff
@@ -85,6 +86,8 @@ const defaultProps = {
   onClick: null,
   onClickAnnotation: null,
   onDoubleClick: null,
+  // ({ subDomain, transformation }) => void
```
evancharlton: Should this be here or on the `propTypes`?

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-154.surge.sh

PR #156, opened 2018-09-13
Copy the loaderConfig data before the series data

evancharlton:
The series object (sourced from props) has the most up-to-date
information about the series (eg: hidden, colors, etc). These should not
be trampled by properties that are stored in the loaderConfig copy.

### DIFF
--- a/src/components/DataProvider/index.js
+++ b/src/components/DataProvider/index.js
@@ -216,8 +216,8 @@ export default class DataProvider extends Component {
     const { loaderConfig, yDomains } = this.state;
     return {
       data: [],
-      ...deleteUndefinedFromObject(series),
       ...deleteUndefinedFromObject(loaderConfig[series.id]),
+      ...deleteUndefinedFromObject(series),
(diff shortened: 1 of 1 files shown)


evancharlton:
As I understand it, this happens because:
1. The loader is passed the series object (via `oldSeries`)
2. The loader then returns that object spread into a new object, but with data
3. This object is then stored in `loaderConfig[series.id]`
Then when a state change happens, we enrich the series again, which:
1. Copies in the data from the new prop
2. Copies in the data **from `loaderConfig[series.id]`**

This means that any changes in the prop-sourced series object will be overwritten.

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-156.surge.sh

evancharlton:
Not yet; I'll try to create a test case in the storybook.

evancharlton:
This is fixed properly with #164, including reproducing it in the storybook, and then fixing the bug.

PR #159, opened 2018-09-20
Make attached areas follow the ySubDomain

evancharlton:
Areas which specify both series IDs and `yval` for start and end should
be scaled with the associated series. This is a regression, introduced
with 36361e40aae.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-159.surge.sh

PR #160, opened 2018-09-20
Have GridLines track ySubDomain rather than yDomain

evancharlton:
This regression was due to missing this in 36361e40


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-160.surge.sh

PR #161, opened 2018-09-20
Fix the liveLoader's use of randomData

evancharlton:
The randomData API changed, and the liveLoader's use of it was missed.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-161.surge.sh

PR #162, opened 2018-09-20
Fix the single-value story

evancharlton:
The randomData loader went through a refactoring and the monoLoader was
missed as part of this cleanup.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-162.surge.sh

PR #166, opened 2018-09-21
Handle publishing over top of prereleases

evancharlton:
When publishing from prerelease to non-prerelease, the loss of a
prerelease version needs to be detected so that the patch version can be
correctly bumped.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-166.surge.sh

PR #167, opened 2018-09-21
Fix PropType warnings

evancharlton:
These props should be set already, but the props aren't populated for
the initial render, so don't require the props to be set.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-167.surge.sh

evancharlton:
Ping.

PR #171, opened 2018-10-01
add onBaseDomainChanged to DataProvider

jasmineleeson:
- add an `onBaseDomainChanged` prop to `DataProvider`
- allow subDomain to be > baseDomain when checking if we are looking at the front of the window
- maintain initial subDomain length in `getSubdomain`

### DIFF
--- a/src/components/DataProvider/index.js
+++ b/src/components/DataProvider/index.js
@@ -51,9 +51,10 @@ const getSubDomain = (baseDomain, subDomain) => {
   if (!subDomain) {
     return baseDomain;
   }
-  const minDomain = Math.max(subDomain[0], baseDomain[0]);
-  const maxDomain = Math.min(subDomain[1], baseDomain[1]);
-  return [minDomain, maxDomain];
+  const subDomainLength = subDomain[1] - subDomain[0];
+  const end = Math.min(subDomain[1], baseDomain[1]);
+  const start = Math.max(end - subDomainLength, baseDomain[0]);
+  return [start, end];
 };
 
 /**
@@ -239,6 +240,9 @@ export default class DataProvider extends Component {
             baseDomain: baseDomain.map(d => d + updateInterval),
           },
           () => {
+            if (this.props.onBaseDomainChanged) {
+              this.props.onBaseDomainChanged(this.state.baseDomain);
+            }
             Promise.map(this.props.series, s =>
               this.fetchData(s.id, 'INTERVAL')
             );
@@ -404,6 +408,7 @@ export default class DataProvider extends Component {
       subDomain: externalSubDomain,
       collections,
     } = this.props;
+
     if (Object.keys(loaderConfig).length === 0) {
       // Do not bother, loader hasn't given any data yet.
       return null;
@@ -518,6 +523,7 @@ DataProvider.propTypes = {
   collections: GriffPropTypes.collections,
   // (subDomain) => null
   onSubDomainChanged: PropTypes.func,
+  onBaseDomainChanged: PropTypes.func,
   opacity: PropTypes.number,
   opacityAccessor: PropTypes.func,
   pointWidth: PropTypes.number,
@@ -529,6 +535,7 @@ DataProvider.defaultProps = {
   collections: [],
   defaultLoader: null,
   onSubDomainChanged: null,
+  onBaseDomainChanged: null,
   opacity: 1.0,
   opacityAccessor: null,
   pointsPerSeries: 250,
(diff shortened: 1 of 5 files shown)


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-171.surge.sh

src/components/DataProvider/index.js
```diff
@@ -51,9 +51,10 @@ const getSubDomain = (baseDomain, subDomain) => {
   if (!subDomain) {
     return baseDomain;
   }
-  const minDomain = Math.max(subDomain[0], baseDomain[0]);
-  const maxDomain = Math.min(subDomain[1], baseDomain[1]);
-  return [minDomain, maxDomain];
+  const subDomainLength = subDomain[1] - subDomain[0];
```
evancharlton: This seems like it will have all kinds of unforeseen consequences. I need to dig into this more to understand what's going to be impacted.

jasmineleeson:
@evancharlton the `baseDomain` is changed internally within `DataProvider` if an `updateInterval` is provided. But you're right that `onBaseDomainChanged` should be called whenever the `baseDomain` is updated manually as well (I didn't realize this was possible).

src/components/DataProvider/index.js
evancharlton: Sorry for taking so long to review this. I put together some test cases (below), but I can't figure out how to make them run properly in griff so I'm just gonna copy the whole thing in here. Anyway, what do you think about the proposed implementation of `getSubDomain`? The current implementation has some known problems, and this PR's version doesn't correctly handle underflowing `subDomain`s.

```
import expect from 'expect';

// make sure that passed subDomain is part of baseDomain
const getSubDomain = (baseDomain, subDomain) => {
  if (!subDomain) {
    return baseDomain;
  }
  const baseDomainLength = baseDomain[1] - baseDomain[0];
  const subDomainLength = subDomain[1] - subDomain[0];
  if (baseDomainLength < subDomainLength) {
    return baseDomain;
  }
  if (subDomain[0] < baseDomain[0]) {
    return [baseDomain[0], Math.min(baseDomain[0] + subDomainLength)];
  }
  if (subDomain[1] > baseDomain[1]) {
    return [Math.max(baseDomain[1] - subDomainLength), baseDomain[1]];
  }
  return subDomain;
};

describe('DataProvider', () => {
  describe('getSubDomain', () => {
    it('handles the base case', () => {
      const baseDomain = [0, 100];
      const subDomain = [25, 75];
      expect(getSubDomain(baseDomain, subDomain)).toMatch(subDomain);
    });

    it('handles when the subdomain goes longer', () => {
      const baseDomain = [50, 100];
      const subDomain = [95, 105];
      expect(getSubDomain(baseDomain, subDomain)).toMatch([90, 100]);
    });

    it('handles when the subdomain goes shorter', () => {
      const baseDomain = [50, 100];
      const subDomain = [45, 55];
      expect(getSubDomain(baseDomain, subDomain)).toMatch([50, 60]);
    });

    it('handles when the subdomain is outside', () => {
      const baseDomain = [50, 100];
      const subDomain = [150, 160];
      expect(getSubDomain(baseDomain, subDomain)).toMatch([90, 100]);
    });

    it('handles when the subdomain is longer than the domain', () => {
      const baseDomain = [50, 100];
      const subDomain = [25, 125];
      expect(getSubDomain(baseDomain, subDomain)).toMatch([50, 100]);
    });
  });
});
```

If you put that into `tests/index-test.js` and run `yarn test` then it should pass. If you want to go for extra credit, you can figure out how to import `getSubDomain` from `DataProvider` and directly test it, as it should be.

evancharlton: yeah you're right -- those don't make any sense as-written. I think what I meant to do was eg: `return [baseDomain[0], Math.min(baseDomain[0] + subDomainLength, baseDomain[1])]`

Basically, the math should never push the bounds of `baseDomain`

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-171.surge.sh

src/components/DataProvider/index.js
jasmineleeson: right, but since `subDomainLength` is guaranteed to be <= to `baseDomainLength` at this point, won't `baseDomain[0] + subDomainLength` always be <= `baseDomain[1]`?

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-171.surge.sh

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-171.surge.sh

PR #175, opened 2018-10-10
Introduce xDomain and xSubDomain

evancharlton:
Instead of using a notion of baseDomain and subDomain, introduce a
xDomain and xSubDomain. These mirror the yDomain and ySubDomain
properties. This change is necessary because, with the addition of
scatterplots, not all charted data has a relevant time property.

This will also let the X values be populated by the extent of the data,
similar to the yDomain is calculated by the range of the data.

The xDomain is still passed into the loader so that it can do the right
thing -- for time-based loaders, the xDomain will continue to represent
time. However, for loaders which have time as an independent axis, they
can manage their own notion of which data to load.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ The outermost component in the hierarchy. The DataProvider is in charge of handl
 
 ```js
 DataProvider.propTypes = {
-  baseDomain: PropTypes.arrayOf(PropTypes.number).isRequired,
+  xDomain: PropTypes.arrayOf(PropTypes.number).isRequired,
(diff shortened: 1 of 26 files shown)


evancharlton:
This looks like a lot of code, but the only bits you really need to review are where you see green lines on the left without matching red lines on the right.

But hey, don't let me stop anyone from going line-by-line.

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-175.surge.sh

PR #176, opened 2018-10-11
Correctly bump the prerelease when bumping

evancharlton:
When publishing the prerelease, if it's a named prerelease, the returned
object will be `[ 'name', 2 ]` -- we need to extract the correct value.

### DIFF
--- a/scripts/publish-release.js
+++ b/scripts/publish-release.js
@@ -43,7 +43,10 @@ function findVersionToBump(currentVersion, versions) {
     .sort((a, b) => semver.patch(b) - semver.patch(a))
     .sort((a, b) => {
       if (semver.prerelease(b) && semver.prerelease(a)) {
-        return Number(semver.prerelease(b)) - Number(semver.prerelease(a));
+        return (
+          Number(semver.prerelease(b).pop()) -
+          Number(semver.prerelease(a).pop())
+        );
       }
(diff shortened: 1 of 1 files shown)


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-176.surge.sh

scripts/publish-release.js
```diff
@@ -43,7 +43,10 @@ function findVersionToBump(currentVersion, versions) {
     .sort((a, b) => semver.patch(b) - semver.patch(a))
     .sort((a, b) => {
       if (semver.prerelease(b) && semver.prerelease(a)) {
-        return Number(semver.prerelease(b)) - Number(semver.prerelease(a));
+        return (
+          Number(semver.prerelease(b).pop()) -
+          Number(semver.prerelease(a).pop())
+        );
```
valerii-cognite: It seems you are mutating array here with `pop` isn't it better just to access a value by index?

evancharlton: I just want the last item in the array, and I don't mind mutating the returned array. (If I use the `length - 1` index, then I have to stash that array somewhere. This seemed cleaner, but I won't fight you if you disagree.)

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-176.surge.sh

PR #177, opened 2018-10-11
Add sub domain control props to DataProvider

jasmineleeson:
- if the prop `isXSubDomainSticky` is set to `true`, the `xSubDomain` state will be updated at every `updateInterval`, similarly to `xDomain`
- if the prop `limitXSubDomain` is provided, it allows the external application to limit the value of `xSubDomain` in  `xSubDomainChanged`, i.e. whenever the `xSubDomain` is manually changed by zoom or pan

### DIFF
--- a/src/components/DataProvider/index.js
+++ b/src/components/DataProvider/index.js
@@ -186,9 +186,6 @@ export default class DataProvider extends Component {
       if (this.props.onSubDomainChanged) {
         this.props.onSubDomainChanged(newSubDomain);
       }
-      if (this.props.onBaseDomainChanged) {
-        this.props.onBaseDomainChanged(this.props.baseDomain);
-      }
       if (this.fetchInterval) {
         clearInterval(this.fetchInterval);
       }
@@ -247,15 +244,15 @@ export default class DataProvider extends Component {
     const { updateInterval } = this.props;
     if (updateInterval) {
       this.fetchInterval = setInterval(() => {
-        const { baseDomain } = this.state;
+        const { baseDomain, subDomain } = this.state;
         this.setState(
           {
             baseDomain: baseDomain.map(d => d + updateInterval),
+            subDomain: this.props.isSubDomainSticky
+              ? subDomain.map(d => d + updateInterval)
+              : subDomain,
           },
           () => {
-            if (this.props.onBaseDomainChanged) {
-              this.props.onBaseDomainChanged(this.state.baseDomain);
-            }
             Promise.map(this.props.series, s =>
               this.fetchData(s.id, 'INTERVAL')
             );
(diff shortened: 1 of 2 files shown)


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-177.surge.sh

src/components/DataProvider/index.js
```diff
@@ -560,20 +562,19 @@ DataProvider.propTypes = {
...
   // xSubDomain => void
   onXSubDomainChanged: PropTypes.func,
-  // xDomain => void
-  onXDomainChanged: PropTypes.func,
   opacity: PropTypes.number,
   opacityAccessor: PropTypes.func,
   pointWidth: PropTypes.number,
   pointWidthAccessor: PropTypes.func,
   strokeWidth: PropTypes.number,
+  isXSubDomainSticky: PropTypes.bool,
```
evancharlton: Could you put some docs on this and `limitXSubDomain`? I (the user) have no idea what they're supposed to do.

src/components/Scaler/index.js
```diff
@@ -53,6 +53,16 @@ class Scaler extends Component {
...
       }
     });
+
+    if (
+      !isEqual(
+        prevProps.dataContext.xSubDomain,
+        this.props.dataContext.xSubDomain
+      )
+    ) {
+      // eslint-disable-next-line
```
evancharlton: Which eslint rule are you disabling here?

evancharlton: (and why?)

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-177.surge.sh

src/components/Scaler/index.js
jasmineleeson: it's the "Do not use setState in componentDidUpdate" rule, which is ok to do as long as you don't introduce an infinite loop

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-177.surge.sh

src/components/Scaler/index.js
evancharlton: Sorry, that was meant to be rhetorical. Please use:
`// eslint-disable-next-line whatever/rule-is-broken`

src/components/DataProvider/index.js
evancharlton: Ping

src/components/DataProvider/index.js
```diff
@@ -416,9 +413,14 @@ export default class DataProvider extends Component {
 
   xSubDomainChanged = xSubDomain => {
     const current = this.state.xSubDomain;
-    if (xSubDomain[0] === current[0] && xSubDomain[1] === current[1]) {
+    const newXSubDomain = this.props.limitXSubDomain
+      ? this.props.limitXSubDomain(xSubDomain)
+      : xSubDomain;
+
+    if (newXSubDomain[0] === current[0] && newXSubDomain[1] === current[1]) {
```
evancharlton: Is this worth using `isEqual(newXSubDomain, current)`?

src/components/DataProvider/index.js
```diff
@@ -592,4 +596,6 @@ DataProvider.defaultProps = {
   yAxisWidth: 50,
   yDomain: null,
   ySubDomain: null,
+  isXSubDomainSticky: false,
+  limitXSubDomain: null,
```
evancharlton: What if we made the default `xSubDomain => xSubDomain`? Then we could skip the conditionals, and it's self-documenting.

PR #178, opened 2018-10-12
Correctly bump the prerelease when bumping (#176)

evancharlton:
When publishing the prerelease, if it's a named prerelease, the returned
object will be `[ 'name', 2 ]` -- we need to extract the correct value.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-178.surge.sh

evancharlton:
Let's just try this again.

PR #179, opened 2018-10-12
Use a correct default xDomain

evancharlton:
When the xDomain prop is not specified, don't default to [0,0] as this
will lead to incorrect values when calculating the extent of the data.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-179.surge.sh

PR #182, opened 2018-10-16
Don't make color a requirement for collections

evancharlton:
Collections don't NEED a color -- they can just fall back to a default.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-182.surge.sh

PR #185, opened 2018-10-17
Use reduce instead of chained filters

evancharlton:
Rather than running filter multiple times on an array, run a single
reduce call.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-185.surge.sh

PR #187, opened 2018-10-18
Clear intervals before setting new ones

evancharlton:
When an interval is set, the old one (if there is one) should be cleared
first. This won't fix all cases of updating while the component isn't
mounted, but if the variable is overwritten, then there's no way to kill
the zombie.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-187.surge.sh

PR #188, opened 2018-10-18
Use filter instead of reduce

evancharlton:
Instead of using a complicated reduce function (introduced in
1f6f672c128), do it with a single filter call. This also fixes a bug
that was introduced with that change where the logic was inverted.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-188.surge.sh

PR #189, opened 2018-10-19
Correct the a logical grouping typo

evancharlton:
This conditional should be false if xSubDomain is not truthy, instead of
incorrectly grouping the truthiness check with the first component of
the conditional test.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-189.surge.sh

PR #190, opened 2018-10-19
clean up clearInterval

jasmineleeson:
- having `startUpdateInterval` after the `await` inside `componentDidMount` means it is possible for `clearInterval` within `componentWillUnmount` to attempt to clear the interval before it is set
- since we now call `clearInterval` at the start of `startUpdateInterval`, there is no need to call it before every call of `startUpdateInterval`


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-190.surge.sh

PR #193, opened 2018-10-26
Use the AxisCollection with Scatterplots

evancharlton:
Now that the Scaler supports independent scaling of the series (across
multiple dimensions), we no longer need to use a unified y-axis to hold
the data. As a result, reuse the `<AxisCollection>` component so that
the logic can be reused.

If a single axis is desired, then the series may be placed into a single
collection, exactly how it is done when using a LineChart.

This change also introduces zooming to the x-axis.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-193.surge.sh

PR #196, opened 2018-10-26
Prevent inverted subdomains

evancharlton:
If the user zooms in very quickly (deltaY <= -1), then the newSubDomain
can become inverted -- the end will be before the start. If this
happens, flip them back around.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-196.surge.sh

PR #197, opened 2018-10-26
Don't snap the ruler to the nearest point

evancharlton:
Instead of snapping the mouse to the nearest point (and returning
potentially incorrect information), always use the *previous* point
since that represents the last-known value at that point in time.

In the future, we should consider returning both -- timestamps included
-- and let the application sort out which one to show to the user.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-197.surge.sh

evancharlton:
Ping

PR #198, opened 2018-10-29
Add support for custom point-rendering

evancharlton:
Allow points to have custom rendering through the `pointRenderer`
option. This property can be set on:

  1. DataProvider
  2. Collection
  3. Series

As with other properties, the most-specific item's property value will
be used.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-198.surge.sh

valerii-cognite:
I believe the function should return either a standard point passed into or override it

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-198.surge.sh

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-198.surge.sh

src/components/Points/index.js
```diff
@@ -53,9 +59,24 @@ const Points = ({
   pointWidthAccessor,
   strokeWidth,
 }) => {
-  const getX = x => boundedSeries(xScale(x));
-  const getY = y => boundedSeries(yScale(y));
-  const points = data.map(d => {
+  if (!drawPoints) {
+    // We should never wind up in this situation, but just in case...
+    return null;
```
valerii-cognite: If `drawPoints` is `false`, it seems instead of `default rendering`, null will be rendered

evancharlton: Which it should, right? `false` should result in no points being rendered.

evancharlton: Oh nevermind -- I see the issue.

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-198.surge.sh

PR #199, opened 2018-10-30
Correctly handle the update events for non-touchmove

evancharlton:
When there are new props for InteractionLayer, we need to update when:

  1. The mouse moves (`mousemove` event)
  2. There was no d3 event (live-loading)

All other reasons need to be ignored (eg, `touchmove` events)

### DIFF
--- a/src/components/InteractionLayer/index.js
+++ b/src/components/InteractionLayer/index.js
@@ -119,6 +119,8 @@ class InteractionLayer extends React.Component {
       const curXScale = xScalerFactory(nextTimeSubDomain, width);
       const ts = prevXScale.invert(touchX).getTime();
       const newXPos = curXScale(ts);
+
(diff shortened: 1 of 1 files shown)


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-199.surge.sh

src/components/InteractionLayer/index.js
```diff
@@ -129,7 +131,8 @@ class InteractionLayer extends React.Component {
       } else if (
         // ruler should follow points during live loading
         // except when the chart is dragging firing touchmove event
-        (((d3 || {}).event || {}).sourceEvent || {}).type === 'mousemove'
+        d3Type === 'mousemove' ||
+        d3Type === undefined
```
valerii-cognite: Initially that was just `!== 'touchmove'`

jasmineleeson: this is changed in #205 anyway

PR #201, opened 2018-10-30
take floor of new subDomain values

jasmineleeson:


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-201.surge.sh

evancharlton:
Why? We can't assume that subdomains end at integer values. For example, a scatterplot might go from [0,1], or some subset thereof.

evancharlton:
In fact: https://griff-react-pr-201.surge.sh/?selectedKind=Scatterplot&selectedStory=Basic%20usage&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

jasmineleeson:
ah of course 🤦🏻‍♀️. But the timeSubDomain does need to have an integer value (the cognite sdk fails otherwise), so can we do a floor within `updateDomains` of `Scaler` for time domains only?

evancharlton:
We could, but it seems like this could also be solved at the app level (and/or at the SDK level)

PR #202, opened 2018-10-31
Handle missing series in the subdomains

evancharlton:
If the series are updated, it's possible to crash because the subdomain
and domain maps in Scaler have not yet been updated. Example logical
flow:

  1. Scaler initialized with series 1.
  2. Scaler's state updated with:
     domainsByItemId: { [1]: { .. } },
     subDomainsByItemId: { [1]: { .. } },
  3. Series 2 added to DataProvider
  4. Scaler notices the new series and updates the state:
     domainsByItemId: { [1]: { .. }, [2]: { .. } },
     subDomainsByItemId: { [1]: { .. }, [2]: { .. } },

When this happens, no other scaling happens in componentDidUpdate
because everthing was reinitialized when there are new series added.

As an additional safety measure, Axes will not crash if an invalid
series is given. Instead, a domain (or subdomain) of `[0, 0]` will be
returned.

### DIFF
--- a/src/components/Scaler/index.js
+++ b/src/components/Scaler/index.js
@@ -11,6 +11,16 @@ import Axes from '../../utils/Axes';
 // the leading edge of the timeDomain.
 const FRONT_OF_WINDOW_THRESHOLD = 0.05;
 
+const containsSameItems = (a, b) => {
+  const reducer = (acc, item) => {
+    if (item.hidden) {
+      return acc;
+    }
+    return { ...acc, [item.id]: item.yDomain };
+  };
+  return isEqual(a.reduce(reducer, {}), b.reduce(reducer, {}));
+};
+
 /**
  * The scaler is the source of truth for all things related to the domains and
  * subdomains for all of the items within Griff. Note that an item can be either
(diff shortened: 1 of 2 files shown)


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-202.surge.sh

src/components/Scaler/index.js
```diff
@@ -59,6 +59,39 @@ class Scaler extends Component {
   }
 
   componentDidUpdate(prevProps) {
+    if (!isEqual(prevProps.dataContext.series, this.props.dataContext.series)) {
+      const domainsByItemId = {};
+      const subDomainsByItemId = {};
+
+      const updateDomainsForItem = item => {
+        domainsByItemId[item.id] = {
+          [Axes.time]: this.props.dataContext.timeDomain,
```
valerii-cognite: seems like `DataProvider` `timeDomain` prop is not Required so it could be not set, shouldn't we set some default in that case?

evancharlton: Done. I also flipped the order of the other domains (and subdomains).

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-202.surge.sh

PR #204, opened 2018-11-01
Let react-sizeme set the size prop for LineChart

evancharlton:
Instead of using the react-sizeme provider and setting the `size` prop
explicitly, use the HOC and have it own the `size` prop.

This prevents a proptype warning but has no functional changes.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-204.surge.sh

PR #205, opened 2018-11-01
Fix mobile panning and ruler behaviour

jasmineleeson:
### Ruler behaviour
- The ruler should stick to the point it was previously set to when:
  - Live loading (i.e. the subDomain changes internally, not due to an action done by the user)
  - Zooming and panning on desktop
- The ruler should not stick to the point it was previously set to, but instead should be aligned with the point closest to the x-position it was previously set to when:
  - Zooming or panning on mobile

### Panning on mobile
Panning on mobile was broken because `this.firstTouch` in `ZoomRect` was `undefined` it had not previously been set in `onMouseMove` (i.e. if the chart had not yet been hovered over or click on). This was implicitly fixed by the fact that we don't actually need to use `firstTouch` because of how the ruler positioning works in `InteractionLayer`

### DIFF
--- a/demo/src/index.js
+++ b/demo/src/index.js
@@ -87,7 +87,7 @@ class App extends Component {
             ruler={{
               visible: true,
               yLabel: point => Number.parseFloat(point.value).toFixed(3),
-              xLabel: point =>
+              timeLabel: point =>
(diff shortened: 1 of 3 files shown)


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-205.surge.sh

PR #206, opened 2018-11-02
Removed the check to see if a ruler exists

danlevings:
This check prevents the `onMouseMove` api from providing data if the ruler is hidden/doesn't exist - breaking opints cursor overview.

Not sure if this code was added to solve some kind of other bug though


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-206.surge.sh

evancharlton:
Seems to work, but I'll let @jasmineleeson tell us where and how this breaks Ruler.

PR #207, opened 2018-11-02
limit timeSubDomain in Scaler

jasmineleeson:
`Scaler` was setting its `timeSubDomain` state before it being limited in `DataProvider`, causing the `timeSubDomain` to jump around (and hence the ruler to jump around) when attempting to pan the chart passed the limit imposed within `limitTimeSubDomain`


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-207.surge.sh

jasmineleeson:
no, since `getTimeSubDomain` and `timeSubDomainChanged` are still called in `DataProvider` (e.g. to set initial state and in `componentDidUpdate`

PR #209, opened 2018-11-04
Add lines on scatterplots

evancharlton:
Through the use of a new `drawLines` prop, allow lines to be rendered on
scatterplots. This is shared between line charts and scatterplots, and
has different defaults -- on line charts, it will default to true, but
it will default to false on scatterplots.

### DIFF
--- a/src/components/DataProvider/index.js
+++ b/src/components/DataProvider/index.js
@@ -275,6 +275,7 @@ export default class DataProvider extends Component {
 
   enrichSeries = (series, collection = {}) => {
     const {
+      drawLines,
       opacity,
       opacityAccessor,
       pointWidth,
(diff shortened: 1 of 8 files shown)


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-209.surge.sh

evancharlton:
(for the peanut gallery) I believe @valerii-cognite was asking about the asymmetry between `drawPoints` (bool | func) and `drawLines` (bool). I think there's an obvious benefit to providing mirroring functionality, but I'd like to avoid replicating it until someone has a use case for it.

katrinetjolsen:
Cool, @evancharlton !

PR #210, opened 2018-11-05
Only show points that are within the timeSubDomain

evancharlton:
Apply a filter before rendering the points so that only points relevant
to the time subdomain (visible time range) are rendered. This allows for
similar behavior to the line chart where changing the time subdomain
shows the low-resolution aggregates before the loader renders the
higher-resolution data.

### DIFF
--- a/src/components/PointCollection/index.js
+++ b/src/components/PointCollection/index.js
@@ -33,7 +33,25 @@ const PointCollection = ({
       Axes.y(subDomainsByItemId[s.collectionId || s.id]),
       height
     );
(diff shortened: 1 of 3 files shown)


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-210.surge.sh

evancharlton:
I'm still unsure about whether I want to merge this (and whether there should be an API to control this). I'll think about it for a day or so and see how I feel.

It just makes the scatterplot-with-context-chart story **really** fun :-)

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-210.surge.sh

PR #216, opened 2018-11-08
better support ruler positioning for multiple series

jasmineleeson:
a few fixes here:
1. don't add a `rulerPoint` if there is no data for that series
1. don't attempt to add a `rulerPoint` if there is no entry in `subDomainsByItemId` for the series. (This can happen when a new series is added but an entry for it has not yet been added to `subDomainsByItemId`)
1. update the ruler position on width change and `subDomainsByItemId` change
1.  don't render `InteractionLayer` if its width will be < 0 (this occurs when sizeMe returns a width of 0 for `LineChart`. this is done because otherwise, the calculated xposition for the ruler will be < 0, causing it to be cleared.

### DIFF
--- a/src/components/InteractionLayer/index.js
+++ b/src/components/InteractionLayer/index.js
@@ -102,22 +102,30 @@ class InteractionLayer extends React.Component {
     const {
       // FIXME: Migrate this to `subDomainsByItemId`.
       timeSubDomain: prevTimeSubDomain,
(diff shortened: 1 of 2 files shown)


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-216.surge.sh

evancharlton:
I can't wait to set ruler free from the confines of InteractionLayer.

Or take it out back and shoot it.

PR #218, opened 2018-11-11
Rework domains & subdomains for x & y axes

evancharlton:
Make x & y domains behave like time: if there is a domain specified,
then the subdomain cannot go beyond those bounds. If neither are
specified, then the subdomain will default to the extent of the data
(with an open-ended domain).

This also removes the notion of a contextSeries. Instead, have the
context chart (if present) simply transform the series to look how it
wants them to look (static domains, drawPoints=false, etc).

This change also introduces an unconsistency: LineCollection takes in a
yScalerFactory with a different signature than xScalerFactory. This will
be cleaned up in a fast-follow PR (to move xScalerFactory to the same
signature as the new yScalerFactory). This refactor was omitted from
this PR to keep it manageable.

### DIFF
--- a/src/components/ContextChart/index.js
+++ b/src/components/ContextChart/index.js
@@ -6,7 +6,6 @@ import LineCollection from '../LineCollection';
 import XAxis from '../XAxis';
 import Annotation from '../Annotation';
 import GriffPropTypes, {
-  seriesPropType,
   annotationPropType,
   scalerFactoryFunc,
 } from '../../utils/proptypes';
@@ -14,6 +13,9 @@ import Brush from '../Brush';
 import AxisPlacement from '../AxisPlacement';
 import multiFormat from '../../utils/multiFormat';
 import Axes from '../../utils/Axes';
+import { createYScale } from '../../utils/scale-helpers';
+import { stripPlaceholderDomain } from '../Scaler';
+import { calculateDomainFromData } from '../DataProvider';
 
 class ContextChart extends Component {
   static propTypes = {
@@ -26,21 +28,20 @@ class ContextChart extends Component {
     xAxisPlacement: GriffPropTypes.axisPlacement,
 
     // These are all provided by Griff.
-    contextSeries: seriesPropType,
-    width: PropTypes.number,
+    domainsByItemId: GriffPropTypes.domainsByItemId.isRequired,
+    series: GriffPropTypes.multipleSeries.isRequired,
+    subDomainsByItemId: GriffPropTypes.subDomainsByItemId.isRequired,
     timeDomain: PropTypes.arrayOf(PropTypes.number).isRequired,
     timeSubDomain: PropTypes.arrayOf(PropTypes.number).isRequired,
-    xScalerFactory: scalerFactoryFunc.isRequired,
     updateDomains: GriffPropTypes.updateDomains.isRequired,
-    subDomainsByItemId: GriffPropTypes.subDomainsByItemId.isRequired,
-    domainsByItemId: GriffPropTypes.domainsByItemId.isRequired,
+    xScalerFactory: scalerFactoryFunc.isRequired,
+    width: PropTypes.number,
   };
 
   static defaultProps = {
     width: 1,
     height: 150,
     annotations: [],
-    contextSeries: [],
     zoomable: true,
     xAxisFormatter: multiFormat,
     xAxisHeight: 50,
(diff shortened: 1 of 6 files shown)


evancharlton:
Just as a hint, the bulk of the review should probably be focused on `Scaler/index.js`

src/components/ContextChart/index.js
```diff
@@ -79,6 +75,15 @@ class ContextChart extends Component {
     );
   };
 
+  getYScale = (s, height) => {
```
valerii-cognite: Shouldn't this be calculated on the `Scaler` level?

src/components/ContextChart/index.js
```diff
@@ -135,10 +139,11 @@ class ContextChart extends Component {
         >
           {annotations}
           <LineCollection
-            series={contextSeries}
+            series={series.map(s => ({ ...s, drawPoints: false }))}
```
valerii-cognite: Seems that drawPoints will override all the user specified behaviour if any provided, is it by design for `ContextChart` not being able to render points?

src/components/Scaler/index.js
```diff
@@ -271,11 +302,13 @@ class Scaler extends Component {
           subDomainsByItemId[itemId][axis] || newSubDomain;
         const existingSpan = existingSubDomain[1] - existingSubDomain[0];
 
-        const limits = ((domainsById || {})[itemId] || {})[axis] ||
-          (axis === String(Axes.time)
-            ? // FIXME: Phase out this single timeDomain thing.
-              this.props.dataContext.timeDomain
-            : undefined) || [Number.MIN_SAFE_INTEGER, Number.MAX_SAFE_INTEGER];
+        const limits = stripPlaceholderDomain(
```
valerii-cognite: Why we need to strip out placeholder here?

src/components/Scaler/index.js
```diff
@@ -83,53 +100,67 @@ class Scaler extends Component {
...
+                ) ||
+                PLACEHOLDER_DOMAIN),
+            ],
+            [Axes.y]: [
+              ...(item.ySubDomain ||
+                stripPlaceholderDomain(
+                  Axes.y(this.state.subDomainsByItemId[item.id])
+                ) ||
+                PLACEHOLDER_DOMAIN),
+            ],
```
valerii-cognite: This one is huge, would you mind to rewrite it into a separate unified function?

src/components/Scaler/index.js
```diff
@@ -83,53 +100,67 @@ class Scaler extends Component {
...
+    );
+    const changedCollections = findItemsWithChangedDomains(
+      prevProps.dataContext.collections,
+      this.props.dataContext.collections
+    );
+    if (changedSeries.length > 0 || changedCollections.length > 0) {
+      const domainsByItemId = { ...this.state.domainsByItemId };
+      const subDomainsByItemId = { ...this.state.subDomainsByItemId };
+
+      []
```
valerii-cognite: Noticed this back and force on `concat` :)

src/components/ContextChart/index.js
evancharlton: The way I see it, no. The ContextChart wants to have non-standard rendering, so it should shoulder that burden

src/components/ContextChart/index.js
evancharlton: Correct -- ContextChart does not render points.

src/components/Scaler/index.js
evancharlton: I worry that splitting this out will actually make it more confusing to read. When they're all side-by-side like this, the repetition is obvious, and you'll notice if something is doing something special.

Thoughts?

src/components/Scaler/index.js
evancharlton: If a domain wasn't specified, it will be given a placeholder value (so that it's a normal array downstream). If this is the case, we want to strip it so that we can go back to the [MIN, MAX] array.

You could argue that instead of using PLACEHOLDER_DOMAIN as the default, maybe we should just use the [MIN, MAX] array, but then it becomes less-obvious that it's a placeholder.

src/components/Scaler/index.js
valerii-cognite: I believe unification is stands for not to observe `something is doing something special`, but I don't have any strong opinion here, so it's up to you.

PR #219, opened 2018-11-19
Correctly determine axes for AxisCollection

evancharlton:
When there are collections which can be on either side of the chart,
AxisCollection needs to handle this correctly. Previously, the following
situation would arise:

Series:
  { id: 1, collectionId: 'left' }
  { id: 2, collectionId: 'left' }
  { id: 3, collectionId: 'right' }
  { id: 4, collectionId: 'right' }

Collections:
  { id: 'left' }
  { id: 'right' }

With this data set, the left axis would have the `left` collection, but
also series `3` and `4`. Similarly, the right axis would have `right`,
`1`, and `2`.

### DIFF
--- a/src/components/AxisCollection/index.js
+++ b/src/components/AxisCollection/index.js
@@ -111,16 +111,11 @@ class AxisCollection extends React.Component {
       filteredCollections.reverse();
     }
 
-    const collectionsById = filteredCollections.reduce(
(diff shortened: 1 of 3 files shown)


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-219.surge.sh

PR #221, opened 2018-11-26
Allow drawing lines without time

evancharlton:
The x domain is not necessarily time anymore, so don't make that
assumption when drawing lines. It's possible that this getTime() call
could be removed entirely, but it's unclear what the side-effects of
doing that would be. Instead, be safe and allow for a smoother
migration.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-221.surge.sh

PR #227, opened 2018-12-18
make ruler work for live loading and sticky subdomain

jasmineleeson:
the overarching problem here was that in `Scaler`'s `componentDidUpdate`, the `subDomainsByItemId` state was being directly mutated due to this shallow copy on line 176:
` const subDomainsByItemId = { ...this.state.subDomainsByItemId };`
This was causing weird behaviour for the ruler in `InteractionLayer`, since it was not picking up when `subDomainsByItemId` was changing. 

The fix to this was:
- not doing a shallow copy of `subDomainsByItemId`
- placing the check for a change in `timeSubDomain` after the check for a change in `timeDomain` in `Scaler`'s `componentDidUpdate` (the new value from `dataContext.timeSubDomain` should be prioritized over a the new value calculated from a new `dataContext.timeDomain`)
- since `subDomainsByItemId` is now being updated properly in `Scaler`, there is no need for `InteractionLayer` to have any knowledge of `dataContext.timeSubdomain`

### DIFF
--- a/src/components/InteractionLayer/index.js
+++ b/src/components/InteractionLayer/index.js
@@ -53,7 +53,6 @@ class InteractionLayer extends React.Component {
     series: seriesPropType,
     collections: GriffPropTypes.collections,
     timeSubDomain: PropTypes.arrayOf(PropTypes.number).isRequired,
-    timeDomain: PropTypes.arrayOf(PropTypes.number).isRequired,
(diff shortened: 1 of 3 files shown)


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-227.surge.sh

PR #231, opened 2019-01-15
Render times when the x-axis is time-based

evancharlton:
When the x-axis is rendering time, render time using the d3 time scale
so that it correctly handles time ticks.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-231.surge.sh

PR #237, opened 2019-01-19
Give grid lines class names

evancharlton:
In order to allow calling applications to provide their own styles, set
class names on the generated grid lines.

Additionally, convert GridLines from an ES6 class to a stateless
function (since the lines don't have any state).


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-237.surge.sh

PR #243, opened 2019-01-20
Convert ContextChart to a stateless function

evancharlton:
ContextChart does not hold any state. As such, it should not be a
stateful ES6 class. Instead, convert it to a stateless function
(supported by helper functions).


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-243.surge.sh

PR #244, opened 2019-01-21
Convert XAxis to a stateless function

evancharlton:
The XAxis component does not maintain any state -- convert it to be a
stateless function.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-244.surge.sh

PR #246, opened 2019-02-04
Lock d3 version to 4.10.0

stonecrafter:
4.12 doesn't play nice with Webpack 4 due to https://github.com/d3/d3-request/issues/24, encountered this issue while upgrading Storybook to v4 (which runs on Webpack 4).
The suggested fix above was to lock the d3 version to immediately before the breaking dependency upgrade on d3-request. In the long term, we should upgrade griff to use D3 v5 which does not use d3-request at all.

Relates to cognitedata/operational-intelligence/issues/1930

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
   "dependencies": {
     "antd": "^3.4.1",
     "bluebird": "^3.5.1",
-    "d3": "^4.12.0",
+    "d3": "4.10.0",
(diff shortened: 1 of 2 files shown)


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-246.surge.sh

PR #249, opened 2019-02-10
Export GridLines as a top-level component

evancharlton:
Instead of controlling GridLines through props, allow it to be added
directly to the hierarchy as a child of the desired chart (eg, LineChart
or Scatterplot). This allows for a flatter prop structure as well as
beginning the migration of business logic from inside of Griff back into
the client application.

### DIFF
--- a/src/components/Scatterplot/index.js
+++ b/src/components/Scatterplot/index.js
@@ -9,14 +9,12 @@ import GriffPropTypes, { seriesPropType } from '../../utils/proptypes';
 import XAxis from '../XAxis';
 import Layout from './Layout';
 import AxisPlacement from '../AxisPlacement';
-import GridLines from '../GridLines';
 import Axes from '../../utils/Axes';
 import AxisCollection from '../AxisCollection';
 import LineCollection from '../LineCollection';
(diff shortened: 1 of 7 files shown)


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-249.surge.sh

src/components/Scatterplot/index.js
```diff
@@ -148,7 +150,19 @@ const ScatterplotComponent = ({
...
+              axes: {
+                ...(child.props || {}).axes,
+                [Axes.x]:
+                  ((child.props || {}).axes || {}).x === undefined
+                    ? String(Axes.x)
+                    : child.props.axes.x,
+              },
+            };
+            return React.cloneElement(child, childProps);
+          })}
```
valerii-cognite: Probably not now, but I would think about having helper for this type of things for `LineChart` and `Scatterplot`

evancharlton: Yeah it will likely be extracted as we get more and more components pulled out like this.

Or maybe they'll diverge and we won't want them consolidated -- time will tell.

PR #250, opened 2019-02-13
Add error handler callback to DataProvider

stonecrafter:
So that other components using this can have custom error handling functionality in case the data fails to come back.

Relates to https://github.com/cognitedata/operational-intelligence/issues/1993

### DIFF
--- a/src/components/DataProvider/index.js
+++ b/src/components/DataProvider/index.js
@@ -457,6 +457,7 @@ export default class DataProvider extends Component {
       y0Accessor,
       y1Accessor,
       yAccessor,
+      onFetchDataError,
     } = this.props;
     const { timeDomain, timeSubDomain } = this.state;
     const seriesObject = this.getSingleSeriesObject(id);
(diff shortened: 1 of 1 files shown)


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-250.surge.sh

src/components/DataProvider/index.js
```diff
@@ -739,6 +745,8 @@ DataProvider.propTypes = {
   // loaderConfig => void
   // called whenever data is fetched by the loader
   onFetchData: PropTypes.func,
+  // Callback when fetching data from backend returns an error
```
evancharlton: Can you document what this signature looks like? Also, it's more of when the loader throws an exception -- we don't know that it's loading from a backend

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-250.surge.sh

src/components/DataProvider/index.js
```diff
@@ -739,6 +745,9 @@ DataProvider.propTypes = {
   // loaderConfig => void
   // called whenever data is fetched by the loader
   onFetchData: PropTypes.func,
+  // error => void
```
evancharlton: What do you think about `(error, params) => void`?

stonecrafter: Haha sorry I think I merged at the same time as you were commenting.
Params as in the ones passed into the loader?

``` 
loaderResult = await loader({
        id,
        timeDomain,
        timeSubDomain,
        pointsPerSeries,
        oldSeries: seriesObject,
        reason,
      });
```

PR #254, opened 2019-02-26
Ensure tickFormatter gets a number (or NaN)

evancharlton:
tickFormatter should always get a number (or NaN if we're somehow trying
to plot a non-number object).


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-254.surge.sh

PR #264, opened 2019-03-24
Convert CollapsedAxis to a functional component

evancharlton:
This component has no state; there's no reason it should be an ES6
class. Instead, convert it to be a functional component.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-264.surge.sh

PR #279, opened 2019-03-28
Prevent a chart from growing continuously in height

danlevings:
When a container is set to 100% height, but the container fails to grow because it has no parent to set its height from, the chart would grow infinitely. This _should_ stop that

### DIFF
--- a/src/components/LineChart/Layout.js
+++ b/src/components/LineChart/Layout.js
@@ -169,7 +169,12 @@ const Layout = ({
     >
       <div
         className="lines-container"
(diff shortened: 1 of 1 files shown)


danlevings:
Fixes https://github.com/cognitedata/operational-intelligence/issues/1615

cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-279.surge.sh

evancharlton:
Hmmm interesting. I thought it was from SVGs having an extra few px due to being block elements, but we'll see! 🤞

PR #287, opened 2019-04-02
Ported Points and PointCollection to TypeScript

evancharlton:
These two come as a pair, so it makes sense to port them at the same
time.

Also move tsc to use `moduleResolution: node` to handle the default
imports/exports.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-287.surge.sh

PR #293, opened 2019-04-16
Build with ES5

evancharlton:
Publish the module as ES5 so that legacy systems can still use the
project.

Also resolve some dependency issues.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-293.surge.sh

PR #296, opened 2019-04-23
Port AxisDisplayMode to TypeScript

evancharlton:
This also moves it to src/utils/ since it's really not tied to the
LineChart component anyway.


evancharlton:
No sense in reviewing this yet; I didn't have `yarn tsc -w` running ...

PR #304, opened 2019-05-05
Explicitly cast domain.map to Domain

evancharlton:
When taking a known Domain (`[number, number]`) and running a `.map()`
function on it, TypeScript loses all knowledge of the fact that this
array is still a `[number, number]` (because it doesn't have to be).
Address this by explicitly casting these back to Domain afterwards so
that we don't need to disable TypeScript checking on the subsequent
line.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-304.surge.sh

PR #305, opened 2019-05-06
Add displayName to all exported components

evancharlton:
When components are exported as a function (such as when using the
Context API), they lose the displayName property. Add this back in so
that external tools (such as storybook's info addon) work better.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-305.surge.sh

evancharlton:
It does -- the issue is that we're actually exporting anonymous functions. If we saved them to a named constant it would work, too (but then we'd have to have the "actual" class/component use a different name.

PR #308, opened 2019-05-06
Don't nest an extra Scaler for Scatterplot

evancharlton:
For some reason, Scatterplot was rendering its own Scaler. This doesn't
do anything except add unnecessary overhead, so go ahead and remove it.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-308.surge.sh

PR #311, opened 2019-05-06
Add an example of how Chart.js can play with Griff

evancharlton:
Griff can be used with other charting libraries, as is demonstrated with
the plotly integration. Show this off with another example integration,
the Chart.js library.

These are super simple examples which show the shape of the data in the
subdomain.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-311.surge.sh

PR #312, opened 2019-05-09
Support having no Series

evancharlton:
Allow charts to exist which don't have any Series. This can happen if
Griff is used in an application where the user can dynamically build a
chart, for example.


cognite-cicd:
[pr-server]
The storybook for this PR is hosted on https://griff-react-pr-312.surge.sh

PR #319, opened 2019-10-31
Compare domains by values to detect changes

jtaaa:
The domain should not be considered changed unless its values have changed. We were previously checking by reference.

This fixes a bug where the domains are recalculated when another series prop (not a domain or subdomain prop) is changed.


jtaaa:
Before:
BrokenDomains
The zoom on each y-axis is reset every time the mouse leaves the axis collection (the series' `yAxisDisplayMode` is updated by the `onMouseLeave` function)

---

Now:
FixedDomains

:))

PR #313, opened 2019-05-30
Refactor the data flow inside of Scaler

evancharlton:
Clean up the internal data flow inside of the Scaler component. This
refactoring had a few main goals:

  1. Fully-own the responsibility of providing scaled data downstream
     alongside the Series and Collection items.
  2. Increase performance by using better checks to detect when the
     incoming Series and Collections objects were externally changed.
  3. Stop exposing `domainsByItemId` / `subDomainsByItemId` to the
     downstream rendering components, in deference to Series and
     Collection objects directly.
  4. Get closer toward enabling per-Series `timeDomains` / `timeSubDomains`
     as documented in #301.

The end result of this is that the (internal) API has nontrivially
changed. Instead of Scaler exposing maps of `domainsByItemId` and
`subDomainByItemId`, it will simply pass through the `series` and
`collections` arrays, but they are now all guaranteed to be
fully-populated with domain and subDomain information. This means that
rendering components can simply render using the `timeDomain`,
`timeSubDomain`, `xDomain`, `xSubDomain`, `yDomain`, and `ySubDomain` domains
(which are guaranteed to be present) and don't need to concern
themselves with how they were populated (user/state, props, calculated
from data, or placeholders).

This does not yet fully enable separate `timeDomain` / `timeSubdomain`
throughout the library, but it removes almost all blockers. The final
piece to this puzzle is in `DataProvider` -- it needs to learn a one more
trick in order to support this (it currently only knows about one
`timeSubDomain`). That cleanup will be saved for a future PR.

PR #328, opened 2019-11-04
Upgrade storybook & dependencies

evancharlton:
Bump the dependencies for storybook and various TypeScript dependencies
so that we can write tests in TypeScript.

This also configures the loader so that the stories can pull the TypeScript
files in directly, instead of from the build folder.


evancharlton:
There's so much that can be done with this (eg: moving all stories to TS) but I wanted to just scope the initial version to "modernize us". (I'm scared of an infinitely-large PR like #313)

PR #334, opened 2019-11-19
[OI-1908][OI-1910] Collection support for areas

jtaaa:
Also fixes some bugs surrounding areas.

Fixes [OI-1908], [OI-1910]

[OI-1908]: https://cognitedata.atlassian.net/browse/OI-1908
[OI-1910]: https://cognitedata.atlassian.net/browse/OI-1910


src/components/InteractionLayer/index.js
```diff
@@ -545,23 +545,29 @@ class InteractionLayer extends React.Component {
 
       let s = null;
 
-      if (a.start.xval) {
```
danlevings: Do we check elsewhere if a.start is defined?

jtaaa: It's required in the PropTypes

src/components/InteractionLayer/index.js
```diff
@@ -545,23 +545,29 @@ class InteractionLayer extends React.Component {
...
-            scaledArea.start.ypos = yScale(a.start.yval);
-          }
-          if (a.end.yval) {
+
+          scaledArea.start.ypos = yScale(a.start.yval);
+
+          if (a.end !== undefined) {
             scaledArea.end.ypos = yScale(a.end.yval);
+          } else {
+            scaledArea.end = {
```
danlevings: So if end isn't defined, it will set it to the bottom right corner? Interesting.
Right now this doesn't seem very consistent. Perhaps instead we should do checks to see if each individual value is defined, and if not, set them to a default?

e.g. instead of checking for .end to be defined, check each pos inside of .end and set them accordingly. and do the same for .start

stories/InteractionLayer.stories.js
```diff
@@ -470,3 +470,86 @@ export const regressionOnMouseUp = () => {
...
+            },
+          ]}
+        />
+      </DataProvider>
+    </React.Fragment>
+  );
+};
+AreasOnSeries.story = { name: 'Areas with no end' };
+
+export const AreasOnSeriesInCollection = () => {
```
danlevings: Does griff storybook have the knobs plugin? Would be super nice to get one of those set up for better testing of this. (or just use normal inputs)

src/components/InteractionLayer/index.js
jtaaa: It sets it to the top right corner. Checking for each position individually could be more flexible. Are you okay with the defaults of start being bottom left and end being top right?

stories/InteractionLayer.stories.js
jtaaa: Yes I added knobs recently! I'm gonna go through all the storybooks, clean them up and sort them into groups better etc in a separate PR after this

src/components/InteractionLayer/index.js
jtaaa: I went ahead and did this ^ 😅. If start is `undefined` you get the bottom left point, if end is `undefined` you get top right and if they are defined but they are missing a `xpos` or `ypos` it gets filled in with the default start or end `xpos` or `ypos`.

danlevings: 🤔 
I would expect something like:
start.ypos === yScale(start.ypos || ySubDomain[0])
start.xpos === yScale(start.ypos || timeDomain[0])
end.ypos === yScale(start.ypos || ySubDomain[1])
end.xpos === yScale(start.ypos || timeDomain[1])

e.g. top left = start, bottom right = end. This seems natural to me atleast

danlevings: This way as well we won't have infinitely sized areas, which will remove the need for setting values to Number.MIN/MAX_SAFE

jtaaa: The changes I've already made use `scaledArea.start.ypos = yScale(area.start.ypos || ySubDomain[0])` and so on. It also checks and sets the whole start coordinate (`xpos` and `ypos`) if `area.start === undefined` and same for end.

My reasoning for using start as bottom-left and end as top-right is that the domain is defined that way already. The first value in an array we pass as a `Domain` is smaller and to the left or bottom. So I think it's more natural to set end to top-right. I can swap it if you think it's very unnatural tho 😅

danlevings: I find it strange but maybe I'm being weird. We should check with some more end users of griff (aka devs) to see what they think. Its a defining part of the API which we don't want to change in the future.

Check in with @evancharlton  @valgussev and @Meemaw to see what they think

jtaaa: @evancharlton suggested we move away from start and end and instead defined `yMin`, `yMax`, `xMin` and `xMax`. That way there's no confusion

jtaaa: I went ahead and implemented @evancharlton's suggestion. This is a breaking change so when we bump Griff in Insight we will have to update how we use areas.
Areas now look like this:
```
{
  id: ItemId,
  seriesId: ItemId,
  color?: string,
  xMin?: number,
  xMax?: number,
  yMin?: number,
  yMax?: number,
}
```

PR #335, opened 2019-11-19
[OI-1988] Add classnames with ids

jtaaa:
Add an id specific classname to each y-axis and area to allow them to be
queried.

Fixes: [OI-1988]

[OI-1988]: https://cognitedata.atlassian.net/browse/OI-1988

### DIFF
--- a/src/components/InteractionLayer/index.js
+++ b/src/components/InteractionLayer/index.js
@@ -566,13 +566,7 @@ class InteractionLayer extends React.Component {
         }
       }
       const color = scaledArea.color || (s ? s.color : null);
-      return (
-        <Area
-          key={`${scaledArea.id || ''}-${scaledArea.seriesId || ''}`}
-          color={color}
-          {...scaledArea}
-        />
-      );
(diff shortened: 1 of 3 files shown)


src/components/InteractionLayer/index.js
```diff
@@ -566,13 +566,7 @@ class InteractionLayer extends React.Component {
         }
       }
       const color = scaledArea.color || (s ? s.color : null);
-      return (
-        <Area
-          key={`${scaledArea.id || ''}-${scaledArea.seriesId || ''}`}
```
danlevings: I'm a little concerned that the seriesId portion here was added for a reason, and might not want to be removed... @evancharlton ?

jtaaa: Merged but because we drop this array in the fragment with a bunch of other arrays of elements at the same level, the area id isn't necessarily going to be unique as it could clash with the annotation ids...

Not sure about why the seriesId might be needed but I'll fix the above issue.

jtaaa: https://github.com/cognitedata/griff-react/pull/337

PR #337, opened 2019-11-20
Lint and fix ts

jtaaa:
Switch to using [@cognite/eslint-config](https://github.com/cognitedata/eslint-config-cognite) and fix all the lint errors that came with the switch.

Also lint `.tsx` and `.ts`, and set the `esModuleInterop` flag so that we can import modules that don't have a default export without the `* as ...`


src/components/Annotation/index.tsx
```diff
@@ -13,14 +12,14 @@ export interface Props {
...
-const Annotation: React.FunctionComponent<Props> = ({
+const Annotation: React.FC<Props> = ({
   data,
   xScale,
   height,
   color = '#e8336d',
   fillOpacity = 0.1,
   id,
-}) => (
+}: Props) => (
```
evancharlton: We shouldn't need `: Props` here

src/components/GridLines/index.tsx
```diff
@@ -53,7 +53,7 @@ const GridLines: React.FunctionComponent<Props & InternalProps & SizeProps> = ({
   width,
   x,
   y,
-}) => {
+}: Props) => {
```
evancharlton: Same thing here (and probably all over this PR; I'll stop commenting it)

src/components/XAxis/index.tsx
```diff
@@ -11,7 +11,7 @@ import { Domain, Series } from '../../external';
 import { DomainsByItemId } from '../Scaler';
 import { withDisplayName } from '../../utils/displayName';
 
-export interface Props {
+export interface Props extends ScalerProps, SizeProps {
```
evancharlton: I think `react-sizeme` actually exposes the necessary types

package.json
```diff
@@ -56,18 +58,28 @@
     "@types/react-dom": "^16.9.3",
     "awesome-typescript-loader": "^5.2.1",
     "axios": "^0.18.1",
+    "babel-eslint": "^10.0.3",
```
Meemaw: Why do we need this?

jtaaa: It's a peer dependency of [@cognite/eslint-config](https://github.com/cognitedata/eslint-config-cognite)

src/components/Annotation/index.tsx
jtaaa: Unfortunately, eslint complains about prop validation since we have the `react/prop-types` rule on. Similar situation to insight

jtaaa:
Tests will pass when #338 is merged into this

src/components/Annotation/index.tsx
evancharlton: Argh.

src/components/XAxis/index.tsx
evancharlton: I don't know if I want components to take in a `size` prop -- that's a pretty clunky API. I think I'd rather just declare our own size prop, honestly.

stories/AxisCollection.stories.js
```diff
@@ -99,37 +99,37 @@ export const events = () => (
...
+  </>
 );
 
 placement.story = {
   name: 'placement',
 };
 
 export const yAxisWidthStory = () => (
-  <React.Fragment>
+  <>
```
evancharlton: We don't need these as fragments, right?

src/components/XAxis/index.tsx
jtaaa: Ah, well that's what we have currently

stories/AxisCollection.stories.js
jtaaa: Nope we don't. Removed

PR #338, opened 2019-11-20
Snapshot test

jtaaa:
This is a follow up to #337.

Add the first snapshot test for a LineChart! This also fixes our jest configuration and removes the complexity of babel.


jtaaa:
Hah! take that Jenkins

PR #340, opened 2019-11-21
Stories and snapshots

jtaaa:
Add some more stories, tests and snapshot tests.

Next step is to make a hook that allows us to update every prop passed to the line chart. We can make snapshots for each prop permutation (or at least a reasonable amount) using these 3 stories... and maybe one more for multiple collections.

Using the hook we can also test how the LineChart updates when we change props which has been an issue before.


package.json
```diff
@@ -18,7 +18,7 @@
     "fix": "gts fix",
     "lint": "eslint 'src/**/*.{ts,tsx,js}' 'stories/**/*.{ts,tsx,js}'",
     "prepare": "yarn compile",
-    "prepush": "yarn lint",
+    "prepush": "yarn lint && yarn test",
```
evancharlton: 😍

jtaaa:
It's not actually a huge change. Most of that diff is the snap

src/storybook/utils/loaders/index.ts
```diff
@@ -0,0 +1 @@
+export * from './loaders';
```
evancharlton: Since we're pulling this out anyway, can we split each loader into a separate file?

jtaaa: Good idea. I didn't actually define any loaders but instead some "loader creator" functions so I renamed the loaders file. Many more loaders will eventually come so it makes sense to separate them.

PR #341, opened 2019-11-22
WIP: Snapshot prop stories

jtaaa:
I thought a lot about it and I think I've chosen a good way to implement snapshot stories.

Instead of using hooks to update the story props we could instead just rerender the component with different props using `react-testing-library`. The problem with re-rendering and snapshot testing against that re-render is that we don't have a visual indication of why the test is failing. The snapshot can get so big that it's impractical to determine what has changed though checking the diff especially because of the huge `svg` components like `path`s that are generated on one line.

By extracting each prop story, we can reference that story to see obvious visual bugs.

So far I have done many of the visual options on `LineCharts` but the rendering of the chart is also dependent on the `DataProvider` as well as the `Collection` and `Series` components within. I'm thinking the props on these components can be tested on `LineChart`s and `Scatterplot`s simultaneously similarly to how we test `SingleSeries`, `MultiSeries` and `MultiSeriesCollection` variations together.

Disadvantages:
* We don't actually get a visual diff. Although we can reference the story we cannot see what the story used to look like, unless we compare against the same story from the base branch.
* A whole bunch of files and stories. Through helper components it is fairly easy to write each story and more can be done to make it even easier. The number of stories does not affect the storybook experience as they are nested away in a folder.
* We aren't testing prop changes. Currently we aren't testing prop changes through snapshot tests since each snapshot is generated from a static story. This will come in the future and will be implemented in their own tests (prop change tests are distinct from prop tests).

Storybook:
Screen Shot 2019-11-22 at 5 50 13 PM


src/components/LineChart/types.ts
```diff
@@ -0,0 +1,126 @@
+import { SizeMeProps } from 'react-sizeme';
+import { ItemId, Domain } from 'external';
+import { AxisPlacement } from 'components/AxisPlacement';
+import { AxisDisplayMode } from 'utils/AxisDisplayMode';
+
+type Point = {
```
jtaaa: These types should be moved to their components eventually.

src/components/LineChart/types.ts
```diff
@@ -0,0 +1,126 @@
...
+  // onClickAnnotation: func,
+  // event => void
+  // onDoubleClick: func,
+  // ({ xSubDomain, transformation }) => void
+  // onZoomXAxis: func,
+  // Number => String
+  // (e, seriesId) => void
+  // onAxisMouseEnter: func,
+  // (e, seriesId) => void
+  // onAxisMouseLeave: func,
```
jtaaa: I haven't typed all the functions yet. The tests I'm creating right now won't use them (though they should be tested separately).

evancharlton:
Time to admit the truth: we'll reopen this later if we revisit this.

PR #345, opened 2019-12-19
[OI-2219] Add grid template row and column values

jtaaa:
Based on scatterplot implementation

Related to [OI-2219]

[OI-2219]: https://cognitedata.atlassian.net/browse/OI-2219


danlevings:
Is this okay to merge @jtaaa ?

PR #350, opened 2020-03-17
Return xAccessor and build completedSeries object in proper order

paulo-menezes:
- xAccessor object is expected to be returned for Operation Support trends chart to work properly.
- completedSeries object was being constructed in a way that domain and subDomain info was being replaced by spread series object. So now all the objects are being spread first and then we set the other properties individually.

### DIFF
--- a/src/components/DataProvider/index.js
+++ b/src/components/DataProvider/index.js
@@ -262,6 +262,12 @@ export default class DataProvider extends Component {
         // Then the global props from DataProvider, if any are set.
         ...dataProvider,
 
+        // Next, copy over defaults from the parent collection, if there is one.
+        ...collection,
+
+        // Finally, the series configuration itself.
+        ...series,
+
         // Then the domains because these are in the DataProvider state, which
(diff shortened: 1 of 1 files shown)


src/components/DataProvider/index.js
```diff
@@ -349,7 +349,7 @@ export default class DataProvider extends Component {
       timeDomain,
       timeSubDomain,
       pointsPerSeries,
-      oldSeries: { data: [], ...seriesObject },
+      oldSeries: { data: [], ...seriesObject, xAccessor },
```
evancharlton: Why do you need this? The xAccessor is passed in from the outside, so the outside app already has a reference to it.

paulo-menezes: Yes, you're right. Just found out how to change it from the OpSup side. Reverting this change.

paulo-menezes:
Managed to adapt OpSup to the latest version of Griff. Closing this PR.

PR #352, opened 2020-07-07
Add ability to recalcuate the data for yaxis if time subdomain changed

ana-gheorghe:
When we are using griff-react (with cogniteloader from griff-loader) in AIR, we are running into an issue where we use the default yDomain provided by griff-react. When we originally load our data, we load it from the span of one year, so many datapoints are aggregated and some of the extreme data points are averaged out. 

However, when we zoom in and adjust the timeSubdomain, we would like to see these extreme data points. But as we change the timeSubdomain, the yDomain doesn't change to reflect the new data (where the extreme datapoints would no longer necessarily be averaged out). The only way we can see these time points is to selectively zoom in/out on the yDomain - which isn't a great user experience.

This can be seen here if you have access to akerbp-test:

https://staging.air.cogniteapp.com/akerbp-test/dashboard/CVX6BTLRndnhkhpDTsIM/system/FGYWEw1VbX5dT9ZSmfca/timeseries/3462954466371299/alert/6648319396225031

With this change - we could provide a prop to DataProvider to tell it to recalculate the data when the timeSubDomain changes.

This feels a bit hacky and like I might be overlooking a better way to do this - so please let me know if there's a better approach you can think of!

### DIFF
--- a/src/components/DataProvider/index.js
+++ b/src/components/DataProvider/index.js
@@ -333,6 +333,7 @@ export default class DataProvider extends Component {
       y1Accessor,
       yAccessor,
       onFetchDataError,
+      recalcDataTimeSubDomainChanged,
     } = this.props;
     const { timeDomain, timeSubDomain, seriesById } = this.state;
     const seriesObject = seriesById[id];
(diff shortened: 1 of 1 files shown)


evancharlton:
So here's the problem: we also need to rewrite our entire CI/CD setup -- we can no longer use Jenkins for public repos.

evancharlton:
Could this also be solved today with the loader returning the new `ySubDomain` to use after fetching new data?

PR #353, opened 2020-07-09
chore: Migrate to GitHub actions for CI/CD

evancharlton:
Change from Jenkins to GitHub Actions for our CI/CD needs.


evancharlton:
I want it to be on `gh-pages` with `/pulls/<pr-number>` but I don't know how to make that happen. We'll deal with the random ID until I figure that out.

Now we're just waiting on the org `NPM_PUBLISH_TOKEN` secret.

PR #354, opened 2020-07-10
fix: context chart resizing issue

mrtraser:
Fixes issue related to `ContextChart` xAxis resizing:

Screenshot 2020-07-10 at 11 31 18

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ es/
 lib/
 npm-debug.log*
 yarn-error.log
+.idea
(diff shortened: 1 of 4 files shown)


evancharlton:
If this is blocking you then you will need to find another way to work around it outside of the `@cognite/griff-react` package as we're unable to push new versions at this time.

mrtraser:
@evancharlton, does it mean that you'll push new version after some time (week, month)?

evancharlton:
> @evancharlton, does it mean that you'll push new version after some time (week, month)?

As soon as we can, yes.

PR #360, opened 2020-10-19
use react-griff-provider

jarlah:


package.json
```diff
@@ -89,7 +89,8 @@
     "semver": "^6.3.0",
     "source-map-loader": "^0.2.4",
     "ts-jest": "^24.1.0",
-    "typescript": "^3.6.4"
+    "typescript": "^3.6.4",
+    "@cognite/react-griff-provider": "0.0.1"
```
valgussev, line 93: wow, thanks a lot of migrating this over as well!

src/components/Annotation/index.tsx
```diff
@@ -26,6 +26,7 @@ const Annotation: React.FC<Props> = ({
     x={xScale(data[0])}
     y={0}
     height={height}
+    // @ts-ignore
```
valgussev, line 29: should we approach this?

src/components/LineCollection/index.tsx
```diff
@@ -93,7 +93,7 @@ const LineCollection: React.FC<Props> = ({
 
 export default withDisplayName('LineCollection', (props: Props) => (
   <ScalerContext.Consumer>
-    {({ domainsByItemId, subDomainsByItemId, series }: InternalProps) => (
+    {({ domainsByItemId, subDomainsByItemId, series }: any) => (
```
valgussev, line 96: Why to remove type?

src/components/Points/index.tsx
```diff
@@ -51,13 +51,15 @@ const Points: React.FC<Props> = ({
...
       }
+      // @ts-ignore
       return +boundedSeries(xScale(func(d, i, arr)));
     });
 
     const [y, y0, y1] = [yAccessor, y0Accessor, y1Accessor].map(func => {
       if (!func) {
         return Number.NaN;
       }
+      // @ts-ignore
```
valgussev, line 62: same here?

src/components/ContextChart/index.js
```diff
@@ -99,7 +98,7 @@ const ContextChart = ({
         s.yDomain,
         Axes.y(domainsByItemId[s.collectionId || s.id])
       ) ||
-      calculateDomainFromData(s.data, s.yAccessor, s.y0Accessor, s.y1Accessor);
+      [0,0];
```
valgussev, line 101: Could you please explain a bit the change here?

valgussev (commented):
Something to keep in mind - griff is open sourced, while the provider you import is not, so such a change will never be able to built.

src/components/ContextChart/index.js
jarlah: Poor (lazy) mans ignore ;) i should have fixed it (like fixing types and stuff) or ts-ignored it. Not remove it lol

src/components/Points/index.tsx
jarlah: Need to rework typings, possibly Export both types and helpers from provider

src/components/LineCollection/index.tsx
jarlah: Because i need to rework the types on the context type

src/components/Annotation/index.tsx
jarlah: Dont remember whats the problem but absolutely

